PR #68, opened 2012-01-11
A handful of changes/fixes we've made over the last couple weeks.

ckohnert:
I'm including the disablement of tapjoy by default, but feel free to not include that particular change.


patrickmeehan:
I wound up moving these over manually (except the tapjoy disabling). Thanks for sending them! :)

PR #196, opened 2012-03-17
[particle-helper] add function atan2rot

tommo:


patrickmeehan:
Thanks, Tommo! Adam also added some addition trig functions to the particle commands. Check them out in 1.0 when you have a second. :)

PR #289, opened 2012-04-18
Fixed an issue with run-host script on Windows for Android host

jlytledistinctdev:
...host script is called on Windows. this seems to have been lost in the shuffle of upgrading the Android host for v1.0 launch.

### DIFF
--- a/ant/README.txt
+++ b/ant/README.txt
@@ -22,3 +22,4 @@
   - Ensure that Apache Ant is properly installed and configured. This includes
     installing a JDK, setting JAVA_HOME and ANT_HOME environment variables and
     adding the Ant bin/ directory to your PATH.
+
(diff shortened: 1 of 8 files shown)


patrickmeehan:
Thanks, Josh!

I'm handing this to Sean to pull in.

jlytledistinctdev:
I also fixed the gitignore file in the xcode/osx folder. It was referring to the old project name so the files weren't getting ignored properly.

jlytledistinctdev:
I just added a quick fix for a typo I found in the moai backwards compatibility lua header file. Turns out as Eric was explaining the file to me I discovered that two of the functions were passing nil variables for the z param instead of an explicit 0. I went ahead and fixed this and added it to my pull request.

PR #321, opened 2012-05-09
Fixing Bug in MoaiDraw

paxtonhare:
This is a pull request for the bug fixed I outlined on the forums here: http://getmoai.com/forums/moaidraw-drawpoints-only-draws-the-1st-point-t740/

The 2 files you need are:
src/moaicore/MOAIGfxDevice.cpp
src/moaicore/MOAIVertexFormat.cpp

The extra files in this pull are from a change I made to allow isometric tile map support. I'm too git challenged to figure out how to just pull the 2 files. Feel free to ignore the other files if you have no interest in iso maps.

--Paxton


patrickmeehan:
Got the fixes, thanks!

Have left the iso stuff for now. I believe the current 'diamond tile' should be already configurable for half-height iso tiles. That was my intent, anyway. Ping me if you want to discuss. :)

PR #418, opened 2012-07-13
Lua-Sqlite3 API and SQL Samples

sgeos:
The Lua-Sqlite3 API has been added to Moai and the OSX, iOS and Android hosts.  I am not in a position to update the other hosts.  I think this API is easier to work with than Kepler Project LuaSQL front end.
http://www.nessie.de/mroth/lua-sqlite3/index.html

The API has a Lua front end that lives in two files (libluasqlite3_loader.lua and sqlite3.lua).  These files were placed in src/lua-modules/moai/.  Perhaps they want to live some place like src/lua-modules/3rdparty/?  package.path was modified on the iOS and Android hosts so require finds the modules.

The Lua-Sqlite3 initialization code is not wrapped in macros that allow it to be disabled.  Perhaps it should be?

An SQL sample has been added for each of the Lua-Sqlite3 and Kepler Project LuaSQL APIs.

AKURunString() has been exposed to the Android host.  It was not ultimately used, but removing the code seems silly.


hach-que:
I've actually wanted sqlite3 support as a 3rd-party module for local data storage, so I'll see if I can get around to tackling the other hosts at some point in time.

sgeos:
It would be fantastic if you could tackle the other hosts for me.  Keep in mind that Moai has had support for SQLite3 via the Kepler Project style LuaSQL API. I added support for the Lua-SQLite3 API, because I think it is easier to use.  See moai-dev/samples/sql in my fork for examples on how to use both APIs.

Kepler Project LuaSQL (in vanilla Moai)
http://www.keplerproject.org/luasql/examples.html

Lua-SQLite3 (in my Fork)
http://www.nessie.de/mroth/lua-sqlite3/index.html

I will be documenting my process soon so that it will be easier for other people to add third party libraries in the future.  The documentation should serve to help you and adding the other hosts should be relatively straightforward.  Here are the steps from memory:
_) Add the 3rdparty/lua-sqlite3 .c and .h files to the build process.
*) Call AKUExtLoadLuasqlite3() in the host (there are a bunch of AKUExtLoad calls).
*) Make sure require "sqlite" can find the lua front end.  This probably requires modifying package.path.  I used AKURunString in the iOS host and modified init.lua in the Android host.  The OSX host uses config.lua.
*) Everything should work, but the various hosts may have quirks.  I had to touch the JNI *and_ write special handling code in the Lua sample to get things working on Android.

sgeos:
Re: hach-que
The documentation can be found here:
https://github.com/sgeos/moai-dev/wiki/Adding-a-Third-Party-Library-to-Moai-SDK

sgeos:
The Lua-SQLite3 API uses the MIT license.
http://www.nessie.de/mroth/lua-sqlite3/index.html
https://github.com/sgeos/moai-dev/blob/master/3rdparty/lua-sqlite3-0.4.1/LICENSE

All of my changes have been released using the CC0 license, so they are in the public domain.
https://github.com/sgeos/moai-dev/blob/master/samples/sql/luasqlite3/COPYING.txt

## 

Contribution changes written in 2012 by Brendan A R Sechter (bsechter [splat] sennue [spot] com)

To the extent possible under law, the author(s) have
dedicated all copyright and related and neighboring rights
to this software to the public domain worldwide. This
software is distributed without any warranty.

You should have received a copy of the CC0 Public Domain
Dedication along with this software. If not, see
http://creativecommons.org/publicdomain/zero/1.0/

Cloven:
I like this call style much better.  Thank you sgeos!

sgeos:
You are very welcome.  I also like this API more and that is why I added it.  The luasql cursors are a pain to work with.

sgeos:
Added external storage support to MOAIAppAndroid.  Also added an external storage sample (unfortunately there is no time to make a really good sample right now).  Perhaps it was bad form to tack this addition onto this repository?

The new additions are also CC0.

makotok:
This pull request or still valid? 
I can not merge anymore. 

This pull request once closed.

PR #431, opened 2012-07-27
Jsfl moai prefix

pygy:
Per request in the forum: http://getmoai.com/forums/flash-samples-suggestion-t1038/

Edit: Another patch crept in...


pygy:
##

PR #438, opened 2012-07-31
CRLF

pygy:
It's been harder than anticipated, but I think I did it...

I've cleaned up the master branch because the other branches diverged too much from one another.

I've tried to merge master into linux, integration, studio and lua-classes (see my repo), but I'm not sure I did everything right, and since the commit is huge, it's hard to tell by looking at the diffs. Here's what I've done:

```
git checkout target_branch
git add -u samples (add all crlf files)
git add -u 3rdparty
git commit -m preCRLF #commit the Windows line ends
git merge master
```

Then resolve the conflicts when they arise (one #ifdef added in the Linux branch and updated Rocket Lobster in integration)

I might have overwritten stuff in the process. I'm trying to compile the various branches before and after the merge, but it takes ages on my machine, so I prefer to submit it as is for you to have a look.


pygy:
I've tried to compile all branches except `linux` (I only have a headless "server" VM right now).

`master` and `studio` build fine. `integration` and `lua-classes` fail in the same way with and without the patch see #439 and #440.

pygy:
It looks like it causes trouble on Windows... The fix seems simple, but I'd rather not do it blind...

http://getmoai.com/forums/visual-studio-newline-problem-t1072/

Could someone have a look?

PR #443, opened 2012-08-01
Lua classes: CRLF

pygy:
See #438


pygy:
Whoops... wrong target.

PR #456, opened 2012-09-06
Adding fill method to MOAIGrid. 

CharlesHolbrow:
I find myself needing to fill MOAIGrid with a single Non-zero value. I like this better than iterating over the grid in Lua or streaming the tiles in. 

I've tested it in my game and with a the deck/tilemap-animated sample. Also tried to break it by using on an MOAIGrid without calling grid:setSize first, but it's working fine.

This is my first contribution to a open source repo, and I'm not exactly sure of the etiquette. Does a change like this warrant an example in the samples folder?

### DIFF
--- a/src/moaicore/MOAIGrid.cpp
+++ b/src/moaicore/MOAIGrid.cpp
@@ -37,6 +37,24 @@ int MOAIGrid::_clearTileFlags ( lua_State* L ) {
 }
 
 //----------------------------------------------------------------//
+/**	@name	fill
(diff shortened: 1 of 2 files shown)


systemrok:
Charles, thanks for contributing. Can you please review and sign the contributors agreement? http://getmoai.com/open-source-contributions.html

CharlesHolbrow:
Okay, docs have been emailed to moaisupport@getmoai.com!

PR #584, opened 2012-10-07
Working luaext support in Linux

dleslie:
In order to get RapaNui to work with Linux I needed to get luaext support working in moai.

Thus, I've added a module to find SQLite3 libs on the host system, a new config var to enable lua ext support, and the necessary CMake build changes to get luaext support working.


dleslie:
Whoops, wrong branch

PR #592, opened 2012-10-18
Log file-open failures

sporksmith:
I added some error messages in cases where a file isn't found or can't be opened. It's nice to know right away when this is the case instead of silent failure and then wondering why an image isn't being displayed, sound being played, etc.

Where possible I tried to check whether the file was actually opened successfully, instead of checking for file existence before trying to open it. The latter approach can miss some corner cases such as if the file has incorrect permissions.

### DIFF
--- a/src/moaicore/MOAIFont.cpp
+++ b/src/moaicore/MOAIFont.cpp
@@ -445,7 +445,7 @@ MOAITextureBase* MOAIFont::GetGlyphTexture ( MOAIGlyph& glyph ) {
 //----------------------------------------------------------------//
 void MOAIFont::Init ( cc8* filename ) {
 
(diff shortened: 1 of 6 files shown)


frantufro:
Hi, could you check if there is more in this pull request that what has been done in https://github.com/moai/moai-dev/commit/37736c94d2b6075ce04c6c43678d9c18fbaa85b0 ?

Did you sign a Contributor Agreement already?

Thanks!

sporksmith:
Yup, 37736c9 only adds logging for failing to load textures. This one does that and also adds logging for failing to load sounds, images, and fonts.

Yes, already sent a contributor agreement.

frantufro:
Amazing!
Thanks a lot!

sporksmith:
No problem. Thanks for merging!

PR #602, opened 2012-12-01
Apsalar renovations

ubershmekel:
Fixed android apsalar having nil mCanonicalUID as default, removed duplicate "&c=" on failure, print error code on failure.

### DIFF
--- a/src/lua-modules/moai/apsalar.lua
+++ b/src/lua-modules/moai/apsalar.lua
@@ -10,6 +10,7 @@
 module ( "apsalar", package.seeall )
 
 require "crypto"
+require "cloud-manager"
 
 --==============================================================
 -- database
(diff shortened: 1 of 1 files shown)


adamvillalobos:
I'm a little concenred with line "require "cloud-manager"" as I don't believe thats part of our default modules.

ubershmekel:
What about all the cloud.encode lines? How do they work?

adamvillalobos:
Ah, the cloud.encode needs to be changed to the url module. Those are a mistake. The rest of the changes are good, except for the one assigning a default Canonical UID. This should be given to you by apsalar, if its not, the services shouldn't be used.

ubershmekel:
I see, thank you. And I have a bit of an OT question - the "DEVICE" global - it isn't something that MOAI sets but it seems to be used in many MOAI standard modules. Is there an alternative for that one as well?

patrickmeehan:
Not planing to support Apsalar moving forward, so closing this.

PR #604, opened 2012-12-05
add INHERIT_VISIBLE and ACTUAL_VISIBLE attr 

tommo:
In a parent/child scene, a node's visibility is determined by parent's visibility and its own visibility.
ATTR_VISIBLE link always overwrites child node's visibile attribute, so it is not useful in this scene.

There's an example in sample/test folder.


patrickmeehan:
Tommo: we've talked about this internally a bit. Is there any reason not to just break the link to ATTR_VISIBLE if you do not want the parent's setting to override the child's?

tommo:
I think it's more common to control visibility in an 'inherit' way , just like a GUI widget system.
Child component ( e.g: a button in a window ) can has its own visibility , while you can control the whole group's visibility through parent ( e.g: the window contains the button ).  
By breaking ATTR_VISIBLE , you just lose the connection between child and parent.

This can be done in script,  so it might be not a must for the host.

patrickmeehan:
Let me mull it over a bit, Tommo. We're going to be focusing on UI systems in the next phase, so I will be thinking about all this stuff again.

tommo:
If you use a tree structure to handle the render stack, parent/child visibility connection is naturally builtin. However, scenegraph in MOAI is a flat structure (which is quite flexible),  parent/child relationships are simulated with attr-link, extra attr-link(either host-side or script-side) is necessary to implement 'tree visibility'.

makotok:
Do you plan to commit this fix?
I look forward to this fix.

patrickmeehan:
Yeah, I'll add this in now. Tommo - thanks for submitting it!

tommo:
Thank you too. lol

patrickmeehan:
OK, I adapted what you wrote and pushed it up to master. I hope I didn't mess it up! If I understand correctly, the child can only be visible if the parent is visible, but the child can toggle its own visibility on or off. So something like:

(parent local)   (child local)  (child result)
1                    1                 1
1                    0                 0
0                    1                 0
0                    0                 0

Is that right?

patrickmeehan:
Also:

ATTR_LOCAL_VISIBLE: the local setting for the prop (read/write)
ATTR_VISIBLE: the final/actual visibility (read only)
ATTR_INHERIT: 'virtual' link just used for inheritance

So we set ATTR_INHERIT to pull from ATTR_VISIBLE.

Connecting ATTR_LOCAL_VISIBLE to ATTR_LOCAL_VISIBLE should give the original behavior.

tommo:
Yes! That's what I wanted.

tommo:
Are you working on UI system now ?

patrickmeehan:
Not just now. Still integrating some Double Fine contributions. Will probably do skinning next and finish our the arcade collision package before tacking UI. But, schedule permitting, should be Soon(tm).

I'm going to close this pull request. Double check my changes when you have a moment and let meknow if they look OK. :)

PR #605, opened 2012-12-06
Fix for MOAIApp not properly included in Chrome build

JDaance:
The lua headers overwrite the MOAIApp created by the chrome source (scons/src/MOAIApp)

The result is MOAIApp being nil in when running lua scripts in a chrome host

This is a quick and easy fix. Perhaps in the long run it should be renamed to MOAIAppChrome and included in the API docs.


adamvillalobos:
I'll go ahead and do the other fix, thanks for pointing this out though.

PR #614, opened 2013-01-10
FMod MusicSystem integration

Paxi1337:
As a part of a course in my master studies we had to implement FMod Wrapper classes for MOAI. Me and my colleague implemented the MusicSystem and some more classes like MusicInfo and MusicEntity that are needed within MusicSystem.
The code itself was compiled and tested as far as we could but is not guarenteed to work since the class MOAIFmodEventSystem is still missing and needed for our class to work.

The includes and classed are included/registered in moaicore if USE_FMODMUSICSYSTEM is defined

It is required by our advisor to do a pull reqest

### DIFF
--- a/src/moaicore/MOAIFmodMusicEntity.cpp
+++ b/src/moaicore/MOAIFmodMusicEntity.cpp
@@ -1,0 +1,30 @@
+#include "pch.h"
+#include <moaicore/MOAIFmodMusicEntity.h>
+
+// register member functions
+void MOAIFmodMusicEntity::RegisterLuaFuncs(MOAILuaState& state) {
+
+	// call any initializers for base classes here:
+	// MOAIFooBase::RegisterLuaFuncs ( state );
+
+	// here are the instance methods:
+	luaL_Reg regTable [] = {
+		{ "getID", _getID },
+		{ "getName", _getName },
+		{ NULL, NULL }
+	};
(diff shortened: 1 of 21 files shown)


frantufro:
Instead of adding this to Moai core, do you think you can create an extension?
moaiext-fmod-music-system ?
You can take a look at other moaiext stuff inside src/ in order to see how this is done.
Thanks!

patrickmeehan:
Good idea, Francisco. And Paxi1337 - much appreciated. :)

Paxi1337:
Sorry for the late reply, I´m currently quite busy but I´ll try to have a look at this next week.
Regards Chris

CircuitCipher:
@Paxi1337 Any updates on this?  It would be a good addition.  Thanks.

jtrunick:
@Paxi1337 I'd love to see this as well. It looks like it might even support mic recording? How can I contact you? I was wondering if its usable, and if you can record from mic with it.

patrickmeehan:
Closing per discussion in reviewers forum. This looks great, but no one has had bandwidth to use it. And we also have two FMOD bindings with a possible third on the way; we don't need four!

If anyone wants to take another look, we'd (at least) want to move this to a module or (better) roll all the current FMOD bindings into one definitive binding.

PR #619, opened 2013-01-22
Initial Camera Support for iOS

antback:
Allows to take a picture with the camera or pick up from the camera roll.

```
MOAIApp.takeCamera(
    function_callback(image_path),
    source,
    x,
    y,
    width,
    height
)
```

**function_callback(image_path)**: It is the function that will be called at the end. The image is copied to the "Documents" directory of the application folder. The function receives the image path as parameter.
**source**: The source use when picking the image. 
The values are
0=UIImagePickerControllerSourceTypePhotoLibrary
1=UIImagePickerControllerSourceTypeCamera
2=UIImagePickerControllerSourceTypeSavedPhotosAlbum

**x**: On iPAD a popover must be used to present the UIPickerController. These are the coordinates and the size of the popover
**y**: y coordinate
**width**: width of popover
**height**: height of popover

Example of use:

```
local img = RNFactory.createImage(Resource:getVignette(0))

img.scaleX = 0.8
img.scaleY = 0.8
img:setOnTouchUp(takePhoto)

function takePhoto(event)
    if (MOAIApp) then
        MOAIApp.takeCamera(function(image_path)
            print ("takeCamera moai function called")
            local i = RNFactory.createImage(image_path)
            sceneGroup:insert(i)
        end,
        1,
        500,
        600,
        10,
        10)
    end
end
```

### DIFF
--- a/src/moaiext-iphone/MOAIAppIOS.h
+++ b/src/moaiext-iphone/MOAIAppIOS.h
@@ -12,6 +12,7 @@
 #import <moaiext-iphone/ReachabilityListener.h>
 
 @class MoaiMailComposeDelegate;
+@class MOAITakeCameraListener;
 
 //================================================================//
 // MOAIAppIOS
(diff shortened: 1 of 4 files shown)


frantufro:
Thanks for this :D

PR #621, opened 2013-01-25
Fixed VSYNC issue on OSX.

clementfarabet:
The default host had no VSYNC on OSX, this patch fixes that.


frantufro:
Thanks a lot!!!

clementfarabet:
Cool thanks for the quick turnaround! Btw, just wanted to add that MOAI is totally awesome; I've been trying quite a few alternatives, and we've settled down on MOAI for good. Will be back with more patches soon ;-).

frantufro:
That's more than amazing!
Don't hesitate on getting in touch with me if you need something!!

PR #622, opened 2013-01-25
VSYNC fix is only for OSX.

clementfarabet:
My previous patch was too coarse: the VSYNC patch was only needed for (and only works on) OSX.


clementfarabet:
Thanks!

frantufro:
Thanks to you for the quick fix :)

PR #643, opened 2013-02-18
Clarify description of textbox color escapes

seebs:
fixing one typo and reducing confusion between number of bits and number of values represented.


frantufro:
Yay! Thanks a lot for helping with documentation! :D

PR #652, opened 2013-02-26
Playnomi

patrickmeehan:
Brought over additions to MOAIImage and fixes to the particle system and pex particle plugins.

### DIFF
--- a/src/moaicore/MOAIImage.cpp
+++ b/src/moaicore/MOAIImage.cpp
@@ -176,6 +176,31 @@ int MOAIImage::_copyRect ( lua_State* L ) {
 }
 
 //----------------------------------------------------------------//
+/**	@name	fillCircle
+	@text	Draw a filled circle.
+
+	@in		number x
+	@in		number y
+	@in		number r
+	@opt	number r			Default value is 0.
+	@opt	number g			Default value is 0.
+	@opt	number b			Default value is 0.
+	@opt	number a			Default value is 0.
+	@out	nil
+*/
+int MOAIImage::_fillCircle ( lua_State* L ) {
+	MOAI_LUA_SETUP ( MOAIImage, "UNNN" )
+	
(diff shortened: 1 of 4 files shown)


src/moaicore/MOAIImage.cpp
```diff
@@ -173,6 +176,31 @@ int MOAIImage::_copyRect ( lua_State* L ) {
...
+	@in		number x
+	@in		number y
+	@in		number r
+	@opt	number r			Default value is 0.
+	@opt	number g			Default value is 0.
+	@opt	number b			Default value is 0.
+	@opt	number a			Default value is 0.
+	@out	nil
+*/
+int MOAIImage::_fillCircle ( lua_State* L ) {
```
frantufro, line 191: Should we change the name to _filledCircle ?

fillCircle makes me think of the action of filling a circle that already exists.

src/moaicore/MOAIImage.cpp
```diff
@@ -173,6 +176,31 @@ int MOAIImage::_copyRect ( lua_State* L ) {
...
+	float r		= state.GetValue < float >( 4, 0.0f );
+	u32 color	= state.GetColor32 ( 5, 0.0f, 0.0f, 0.0f, 0.0f );
+	
+	self->FillCircle ( x0, y0, r, color );
+	return 0;
+}
+
+//----------------------------------------------------------------//
 /**	@name	fillRect
 	@text	Fill a rectangle in the image with a solid color.
```
frantufro, line 205: The same for this one.

src/moaicore/MOAIImage.cpp
patrickmeehan, line 191: Was following the semantics of MOAIDraw. I tend to prefer verbing nouns, so would go with drawFilledCircle () if anything. But fillCircle is succinct.

PR #657, opened 2013-03-03
Remove 1 character to make android host work.

eiriksm:
I don't know how it ended up there, but the character "Ï" will prevent this import from working :)


eiriksm:
Just now saw that this is fixed in the keyboard branch, sorry!

ConghuZhao:
test

PR #662, opened 2013-03-06
Adding Lua 5.2 support to Windows and Android build systems.

selitvin:
Added switch to VS 2010 and Android build to switch between Lua 5.2 and Lua 5.1 interpreter.

When building with 5.2, luaext is not included in the build because since migration to 5.2 of this module is not trivial. Also some methods in MOAILuaRuntime are disabled for the same reason.

Since luasocket is not built with Lua 5.2, ZeroBraneStudio debugger won't be able to connect to the MOAI host.


patrickmeehan:
Hey, Yevgeni - I'm holding off, but haven't forgotten this. I will look into Lua 5.2 (and LuaJIT) after we've explored switching over to cmake for IDE project generation. We're getting to the point where there are enough configuration options that piecemeal integration with all of the IDE projects is not really manageable. Will come back to this contribution (the code changes in particular) once we have a handle on the new way of doing the build.

ylitvin:
Cool. Since most of the changes are build system related, I will reimplement the support for cmake once it becomes available.

PR #664, opened 2013-03-08
fix particle helper

tommo:
fix a reg allocation bug; 
add support for multiple return value to support function 'angleVec' ( not useful since we have sin/cos );


frantufro:
Tommo, can you rebase this and send it over?
It can't be automatically merged.

PR #670, opened 2013-03-13
Fire EVENT_TIMER_END_SPAN a bit later

superquadratic:
I feels more natural if the event is fired after the action's time and cycle have been adjusted and EVENT_TIMER_KEYFRAME is fired (if needed).

See discussion on the forum: http://getmoai.com/forums/timing-of-event-timer-end-span-t1713/


superquadratic:
Screwed up commit message, will create a new pull request.

PR #679, opened 2013-04-09
MOAIImage:convertToGrayScale ();

btatarov:
Example usage:

image = MOAIImage.new()
image:load("moai.jpg")
image:convertToGrayScale()

texture = MOAITexture.new()
texture:load(image)

sprite = MOAIGfxQuad2D.new ()
sprite:setTexture ( texture )
sprite:setRect ( -128, -128, 128, 128 )

base = MOAIProp2D.new ()
base:setDeck ( sprite )
base:setLoc ( 300, 300 )

myLayer:insertProp ( base )

### DIFF
--- a/src/moaicore/MOAIImage.cpp
+++ b/src/moaicore/MOAIImage.cpp
@@ -566,6 +566,21 @@ int MOAIImage::_writePNG ( lua_State* L ) {
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


romainmoreau:
A ready to use solution is to add a shader on the prop.

PR #682, opened 2013-04-15
Fix input lag with the iOS host

Kaali:
iOS host has severe input lag if the frame rate dips below the target and accelerometer is enabled. This can result in a lag of seconds.

This pull request does a bit of an ambitious change, and puts the renderer on its own thread so that the input can get through unblocked. A semaphore blocks the update loop so it won't run if the renderer is drawing.

Another input source, that was previously blocked, is the MOAIKeyboardIOS which can send events to Lua regardless of the semaphore, and can result in a crash if the Lua code changes renderable stuff, such as text. The other commit in this pull request queues input from the keyboard, and pushes them to the listeners at the proper place to prevent crashes.

There probably are other places which might crash the renderer with out of sync changes; which makes this change a bit dangerous, but I still feel it is a necessary one as the input lag can be really severe.

### DIFF
--- a/src/aku/AKU-iphone.h
+++ b/src/aku/AKU-iphone.h
@@ -34,5 +34,6 @@ void			AKUNotifyRemoteNotificationRegistrationComplete	( NSData* deviceToken );
 void			AKUSetConnectionType							( long type );
 void			AKUViewDidRotateFromInterfaceOrientation		( UIInterfaceOrientation orientation );
 void			AKUSetFrameBuffer								( GLuint frameBuffer );
+void			AKUProcessKeyboardEventQueue							();
 
 #endif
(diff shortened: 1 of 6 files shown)


jeiting:
The proper fix to this is probably to reduce the rate at which you are sending accelerometer events to the Moai context. What is your update rate on the accelerometer?

Kaali:
It is at the default rate from the Moai iOS sample project, 60 fps. Actually we removed accelerometer from our project, but these commits are still in place and haven't caused any problems.

Kaali:
Might result in a crash when backgrounding. Haven't looked into it yet.

superquadratic:
The preferred fix is to reduce the accelerometers update interval in MoaiView.mm.

PR #684, opened 2013-04-19
Draft of changes to add color/scale to grids.

seebs:
Okay, I am not sure whether this pull request will be correct now. If it has a lot of PenColor stuff, it's the previous one which was buggy. If it has a samples/grid/grid-color-scale, it's legit.

This is an update which adds the ability to specify RGBA color and scale per-item in grids. I am not sure whether I caught all the loose ends. It does not handle serializing. It's intended to be used for the Miracle Modus app, where I do a lot of things with large grids of squares, and they're painfully slow right now in Corona, and also painfully slow in Moai if I write the vertex buffers myself, computing them in Lua. Adding the color option seems to produce the results I want.


patrickmeehan:
This is interesting. Will check it out and bring it over soon. Have had some similar thoughts myself, so glad to see you went ahead with it!

patrickmeehan:
Hey, so this looks really good. I'm pretty sure that adding 'color' to the gfx device write quad won't cause any performance hit... I think the vertex format has a color component anyway. I will verify that.

Do me a favor... I just made a few breaking changes to master. Once all the projects have been fixed, I will make a separate branch for this and ask you to re-submit the pull request pointing at that branch. (I will never understand why github doesn't let me change that myself. Or maybe it does?)

seebs:
From an email thread:

---

On Wed, Apr 24, 2013 at 5:16 PM, Seebs seebs@seebs.net wrote:

On 24 Apr 2013, at 19:11, Patrick Meehan wrote:

 Hey, so this looks really good. I'm pretty sure that adding 'color' to
the gfx device write quad won't cause any performance hit... I think the
vertex format has a color component anyway. I will verify that.

It did. I didn't change whether a color was written, I just (optionally)
changed what it was; it was previously always WriteFinalColor4b(), which
was the penColor \* ambientColor. So the sole change is that, if a color
argument is present for TransformAndWriteQuad, we get:

```
            // left top
            this->Write ( vtx [ 0 ]);
            this->Write ( uv [ 0 ]);
            this->WriteFinalColor4b ();
```

turned into
                // left top
                this->Write ( vtx [ 0 ]);
                this->Write ( uv [ 0 ]);
                this->Write < u32 >( color32 );
with, up at the top:
        USColorVec computed;
        u32 color32;
        computed.mR = this->mFinalColor.mR \* color.mR;
        computed.mG = this->mFinalColor.mG \* color.mG;
        computed.mB = this->mFinalColor.mB \* color.mB;
        computed.mA = this->mFinalColor.mA \* color.mA;
        color32 = computed.PackRGBA ();

And the reason it's done this way is so the determination of whether to do
that computation is decided at compile time, and things which aren't using
the color won't do the extra work.

 Do me a favor... I just made a few breaking changes to master. Once all
the projects have been fixed, I will make a separate branch for this and
ask you to re-submit the pull request pointing
at that branch. (I will never understand why github doesn't let me change
that myself. Or maybe
it does?)

I am pretty sure it _does_ allow it, I just don't know how. But sure, I
can re-apply that on a new branch. And probably also fix the serialize.

Note: I note another conceptual question about serialize, which is that it
doesn't appear to check the size of the serialized data. So if you change
the size of the grid, I'm not sure what happens. It might make sense to

## have it check.

On 23 Apr 2013, at 15:21, Patrick Meehan wrote:

Will double check the serialization - will want to support that when I
bring this over.

Yeah. I just didn't have a clear idea on how to do it. I am sort of tempted by the notion of adding HAS_COLOR/HAS_SCALE flags, but I concluded that the tests would be more expensive than the multiplication for the scale case, at least.

My first draft tried to use u32 colors, but it turns out to be significantly more expensive and/or harder to multiply those.

Have been thinking about exposing visibility and near-visibility as well...
you are anticipating quite a few things that need to be added. Looking
forward to seeing what you do! :)

Biggest things I want that I don't see right now are ways to get native buttons and text fields. It is specifically the nativeness I need, not the appearance -- I would love to make software that can be used by people who are relying on assistive technologies, and prerendered pixels of characters can't be seen by screen readers. iOS in particular has a fairly large pool of users who can't see, or can't see very well. That one's probably too hard for me to do, though.

It might not be a bad idea at all to have a utility function which takes a lua state, and an argument index, and returns a uscolorvec based on the 4/3/2/1 following arguments, so there'd be a consistent convention in the API for color specifications, and a consistent implementation of it.

-s

pkulchenko:
> I will make a separate branch for this and ask you to re-submit the pull request pointing at that branch. (I will never understand why github doesn't let me change that myself. Or maybe it does?)

Yes, but not through the UI. You should be able to pull the changes to any branch you want when you do this manually (you can see this when you click on the "command line" link next to the "merge pull request" button):

  git checkout -b mybranch master
  git pull https://github.com/remote-repo.git remote-branch
  git checkout master
  git merge mybranch

It will also close the pull request, but only after you merge your changes back to master.

I do it this way most of the time anyway as I test changes before merging them.

Just be careful with what you are pulling as (technically) the branch you are pulling from may have other changes.

seebs:
While I'm thinking about redoing the grid stuff: I am toying with the
notion of adding a Rotation also. In Corona, the things roughly equivalent
to props have a fill color, a scale, and a rotation. It occurs to me that
it might not be particularly hard to add a rotation option to grid items.
(Like, individual items, not just the grid as a whole.)

So, questions:
1. What are the breaking changes and when are they expected in so I should
rework my patch and make sure it still works?
2. How would you feel about adding a setDrawCallback() to a prop which
works roughly like the scriptDeck one, but works on generic props? I think
I can do this; my plan would probably be to put the actual invocation in
MOAILayer::Draw.
3. I note that MOAILayer::Draw and MOAIProp::Draw both have
UNUSED(subPrimID) but then use it.
4. By the time I get as far as a scale and rotation, I wonder if I
shouldn't be just giving the grid a collection of MOAITransform2Ds and
having it apply them per square.

... BTW, MOAITransform2D::addRot() does nothing and takes the wrong number
of arguments. :) Come to think of it, a lot of stuff in there looks to do
nothing; is that not actually in use anymore, or is it just unfinished?
Hmm. It's all #ifdef DOXYGEN.

The hello-moai sample convinces me that there must be somewhere a
moveRot() function which, given two arguments, does a z rotation lasting
for the duration specified by the second argument, but I can't figure out
where that is. (... followup: It's in moai.lua over in lua-headers.)

seebs:
Just an update: I now understand the transform and setAttrLink stuff a lot better, so I may be redoing some of this with that. But!
As of last I checked, I can't actually build moai-dev on OS X. So I can't rework things against master right now.

CircuitCipher:
@seebs Can we close this in favor of #777?

CircuitCipher:
From IRC:
[16:49:38] seebs     Oh, wow, was the old pull request for the grid stuff still around? If so, yes, the old one from months ago can totally go away.
[16:50:03] joshua_jandyco    okay thanks seebs

PR #691, opened 2013-05-03
New DISABLE_TWITTER macro to not compile MOAITwitterIOS.mm so that the main app doesn't have to link against the TW framework.

JaviSoto:


CircuitCipher:
Closed with the merge of PR #763

PR #694, opened 2013-05-09
Fix for error messages when setting bounds to nil

alexey-dc:
According to the docs, settings bounds to nil is supported: http://getmoai.com/docs/class_m_o_a_i_prop.html#ac4e79263e7888512d117368142a499d7

This works fine, however it gives a stack trace as if an error has just occurred. This patch removes that stack trace printout.


frantufro:
Hey Alexey, thanks for this!
Can you make the pull request point to development instead of master?
Thanks!!!

PR #695, opened 2013-05-10
Fixing iOS build after reorganization

kibiz0r:
Some sources in moai-iphone looked like they should be in separate build targets, such as MOAIFacebookIOS => libmoai-ios-facebook, but some seem like iOS-specific additions to targets; like MOAIGameCenterIOS => libmoai-ios, or MOAIHttpTaskNSURL => libmoai-ios-http-client.

I split those sources into separate targets where it made sense, and ended up with a set of targets like this:

Xcode targets

What is the role of libmoai-ios-3rdparty supposed to be? Is it supposed to be a "give me all of the 3rd-party stuff I might want" target, or just the basics that are necessary for using Moai?

Also, I'd like to generate as much of this as possible through CMake, but I have limited experience writing CMake stuff, so maybe @franciscotufro could help with that.

@patrickmeehan

### DIFF
--- a/src/host-glut/GlutHost.cpp
+++ b/src/host-glut/GlutHost.cpp
@@ -5,6 +5,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <moai_config.h>
+#include <lua.h>
 #include <lua-headers/moai_lua.h>
 #include <host-glut/GlutHost.h>
 #include <string.h>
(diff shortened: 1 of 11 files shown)


patrickmeehan:
Yep, 3rd party is all of the dependencies.

kibiz0r:
Merged into development.

PR #698, opened 2013-05-10
Adding Lua 5.2 support to Windows and Android build systems

selitvin:
Added switch to VS 2010 and Android build to switch between Lua 5.2 and Lua 5.1 interpreter.

When building with 5.2, luaext is not included in the build because since migration to 5.2 of this module is not trivial. Also some methods in MOAILuaRuntime are disabled for the same reason.

Since luasocket is not built with Lua 5.2, ZeroBraneStudio debugger won't be able to connect to the MOAI host.

Built on top 1.3 version, so the integration into current build system might not be trivial.


pkulchenko:
Yevgeni, luasocket can be compiled with Lua 5.2, but one would need to use [unstable branch](https://github.com/diegonehab/luasocket/tree/unstable) from the github repo. I also have a [fork](https://github.com/pkulchenko/luasocket/tree/lua52-mingw) of it to add support for compiling it on Windows using mingw (tested with 5.2); here is the relevant [pull request](https://github.com/diegonehab/luasocket/pull/39).

selitvin:
For internal Tango builds I ported the entire moai luaext to 5.2. I can not be 100% sure I did not introduce any bugs, but I don't remember doing anything criminal while porting. 

I was able to use ZeroBrane Studio and connect to 5.2 based Moai. I did not have time to organize it into a clean pull request. Hopefully will have time to clean up that code and submit a separate pull request for that later.

patrickmeehan:
Thanks. Yevgeni! Will work on this today and over the weekend. :)

PR #701, opened 2013-05-14
Improved compatibility issues MOAIFont.

makotok:
Sample of the current MOAITextBox not work all.
By performing this fix, would like to work as before.


makotok:
All right.
I submit again.

PR #702, opened 2013-05-14
MOAIProp - isVisible property (getter)

btatarov:


frantufro:
If you want to avoid the extra commits you can always use git commit --amend, it will reopen the last commit and you can add the new stuff to it.

PR #706, opened 2013-05-19
Enabled popen on OSX to enable unit testing via "busted"

Odie:
Hi,

Unit testing via "busted" was not working on OSX because io.popen was not enabled. To reproduce the bug, simply run the sample in moai-dev/samples/test/popen on OSX. The sample bails out indicating that 'popen' is not supported.

This happens because no compilation flag was included to enable popen. See 3rdparty/lua-5.1.3/src/luaconfig.h:668. The fix is to add the LUA_USE_POPEN when compiling on OSX.

Actually, it appears that on OSX, the proper flag to include is LUA_USE_MACOSX, which will also enable a number of other flags. I wasn't sure if the exclusion of the LUA_USE_MACOSX flag was by design, so the fix was made by adding LUA_USE_POPEN instead.


superquadratic:
I think `LUA_USE_MACOSX` is fine. New pull request: #758

PR #709, opened 2013-06-02
Improved visibility settings

makotok:
Improved operation of the display.
- Improved the problem of reflection of the display is delayed.(# 604
  Bug)
- Set immediately after the SetVisible, to solve the problems that
  ATTR_VISIBLE does not reflect.
- Enable the Visible of MOAIParticleSystem.


makotok:
Recently, merge speed is slow.
Did you have something?

PR #711, opened 2013-06-04
Html Host (emscripten)

halfnelson:
Html5 js host based on emscripten and a custom written untz wrapper using howler.js and custom host code.

It is based off your current master but you probably don't want to merge with that. The cmake files have been changed as has one of the samples. Since you were keen to have a look I am submitting this now before I clean up the changes that didn't need to be made (hello-moai sample, and libcrypto stuff in zlcore?). Instructions for building are in the readme.md files in the host-html folder under cmake and in the host-template folder under src/host-html.

Have fun :) 

> Tested on linux only (building that is, it runs on chrome and firefox on all platforms and on chrome for android (with webgl enabled in settings)


frantufro:
I'm so sad to say this, but this won't work right now, do you think you can rebase with development? Since the whole SDK changed a lot from master, we're close to a major release and there are huge structural changes :(

halfnelson:
yeah np. I'll sleep first, I'll give it a go tomorrow.

frantufro:
haha! great! Thanks a lot!

PR #712, opened 2013-06-05
Html5 js host support

halfnelson:
Latest HTML5 JS host.

Rebased to latest development. I'm not sure how your cmake was working with the current setup but using a copy of your blackberry build script the paths in the third party libs were traversing back one level too far. I tried to keep those mods in a different commit so that you could skip it easily. 

Just watch out for the freetype cmake file, that has been rewritten so as not to build/export things twice (h files include c files!). 

A little bit lazy with my GL Enum remapping, just defined BUILD_NACL to pick up the enum mappings.

I have excluded glew from third-party for html build but It was done differently to how you did it for blackberry. Feel free to change it.

I stub out all of untz sources with a webversion for the untz lib. This is done with a big `IF` block in the cmake file for untz. Maybe it would be better to be a different cmake file.

The readme's will need to be updated, I'll do that later.


JDaance:
What are the obstacles to merging this? Does it mess a lot with other builds?

halfnelson:
HA, looks like the pull request went a bit crazy and is tracking my dev branch
Don't merge this. It changes heaps of things.It might be possible to recreate the pull request for just the Html5 host stuff, but even though it is working, it hasn't been fully tested and only builds on linux. I will keep working on it and resubmit later.
 It was submitted as the ZL guys were going to help test it. I don't think they have time and the branch has diverged

PR #713, opened 2013-06-06
Adding twitter functionality to Android host.

CircuitCipher:
Adding twitter functionality to Android host using Twitter4J library.


CircuitCipher:
I realize with the re-organization, this request is not quite pullable.  When the release is over, let me know where to pull against and I will update the request.  Until then, this can be used as an idea of what I am proposing.  It is currently working in my own private repo that is staying at the early April version of the repo.

CircuitCipher:
Yes,  thanks for catching that.  I'll probably be regenerating this pull request against development branch with that fixed and a few other things.

PR #728, opened 2013-07-18
A necessary switch to freeglut (multi-touch feature)

atis--:
Moai uses GLUT which, according to the _freeglut_ web page http://freeglut.sourceforge.net/, has not been updated in over a decade. That, by itself, may not be reason enough to switch to freeglut. However, I believe there is a **real and immediate benefit** to doing so!

Not only is freeglut a drop-in replacement for GLUT, in its later releases it adds Windows **multi touch support** (WM_TOUCH). This is great news for those who happen to have a touch sensitive monitor under Windows (and possibly other desktop systems as well). One can then develop touch-enabled applications directly on the desktop (faster prototyping).

This pull request adds code to src/host-glut that employs freeglut's new multi-touch functionality. Switching all the project files properly from GLUT over to freeglut is beyond my competence, but code-wise nothing else needs to change: **freeglut just works**.

Regards,
Atis

### DIFF
--- a/src/host-glut/GlutHost.cpp
+++ b/src/host-glut/GlutHost.cpp
@@ -53,7 +53,7 @@
 
 #ifdef _WIN32
 
-	#include <glut.h>
+	#include <freeglut.h>
 	
 	#if MOAI_WITH_FOLDER_WATCHER
 		#include <FolderWatcher-win.h>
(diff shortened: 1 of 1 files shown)


makotok:
This modification results in an error in the non-Windows OS.
code that depends on the freeglut exists.

```
glutMultiButtonFunc ( _onMultiButton );
glutMultiMotionFunc ( _onMultiMotion );
```

In addition, I want you to re-request to the development branch.

PR #738, opened 2013-08-22
Contribution Guidelines

superquadratic:
I think it's great to have a bunch of community maintainers now and I'm excited to see some active development. Before redefining the world of mobile game develoment as we know it, I thought it would be a good idea to agree on some guidelines, so we're all on the same page as to how we want to work. :)

This is my proposition, in the form of a [CONTRIBUTING.md](https://github.com/blog/1184-contributing-guidelines) file. As always, there are exceptions to the rule, and nothing is set in stone. Let me know what you think, what I forgot, or if we need these kind of guidelines at all.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,18 @@
+# Moai Contribution Guidelines
+
+Contributions to the Moai SDK are greatly appreciated. We hope that these guidelines make contributing and accepting contributions as easy as possible for everyone:
+
+## For Contributors
+
+* [Fork the repository](https://github.com/moai/moai-dev/fork), create a branch, commit your changes, and open a pull request.
+* Pull requests with bugfixes should be opened against the master branch, while features should go into the development branch.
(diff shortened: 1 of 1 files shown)


CircuitCipher:
Great idea and this looks good to me.  I'll mull it over and see if I think of anything else to add.  

Edit:  Looked it over more and thought about it some.   Maybe we can add some information encouraging contributors to create samples if they are adding new functionality as this is helpful for those looking to use the code.

makotok:
I think it is good.
There is a point to be worried about some.

1)
The merged to master, we should merge immediately to development?

2)
How to be if someone want to be a team member?

superquadratic:
> 1)
> The merged to master, we should merge immediately to development?

Good point. The master branch needs to be merged into development regularly. Ideally, we should do it every time we merge something into master. I added a bullet point to the list.

> 2)
> How to be if someone want to be a team member?

I'm not sure, but we should definitely encourage people who have a track record of contributions to join the team.

Issue #723, opened 2013-06-27
Crash with flower library samples' main.lua


virtualritz:
I'm using version-1.4p0-173-g553dad6 from github. I build the executable following the steps outlined [here](http://getmoai.com/wiki/index.php?title=Building_Moai_From_Source_With_Xcode) (but I also tried the executable produced by running [build.sh](https://github.com/moai/moai-dev/blob/master/xcode/osx/build.sh), same result).

When I to run the flower library's (V2.1.1-71-g9923dbd) samples' [main.lua](https://github.com/makotok/Hanappe/blob/master/projects/flower-samples/src/main.lua), I get this crash:

```
Process:         moai [25634]
Path:            /Users/USER/*/moai
Identifier:      moai
Version:         ??? (???)
Code Type:       X86 (Native)
Parent Process:  lua [23775]

Date/Time:       2013-06-27 16:01:57.600 +0200
OS Version:      Mac OS X 10.7.5 (11G63b)
Report Version:  9

Interval Since Last Report:          983195 sec
Crashes Since Last Report:           -10
Per-App Crashes Since Last Report:   2
Anonymous UUID:                      2CB52CBC-E7E1-423C-9C09-1EF17BF0A5E7

Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Codes: KERN_PROTECTION_FAILURE at 0x0000000000000014

VM Regions Near 0x14:
--> __PAGEZERO             0000000000000000-0000000000001000 [    4K] ---/--- SM=NUL  /Users/USER/*
    __TEXT                 0000000000001000-000000000044c000 [ 4396K] r-x/rwx SM=COW  /Users/USER/*

Application Specific Information:
objc[25634]: garbage collection is OFF

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   moai                            0x000b3b0d std::_Rb_tree<unsigned int, std::pair<unsigned int const, MOAIGlyph>, std::_Select1st<std::pair<unsigned int const, MOAIGlyph> >, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, MOAIGlyph> > >::_M_begin() + 13
[...]

PR #740, opened 2013-08-25
Merge development into master

superquadratic:
For the two builds that I use regularly, the development branch is more stable than master at the moment:
- **iOS:** The Xcode project is broken on master, on development it is fixed.
- **OS X:** The OS X host from master crashes when running our game (and the flower samples, #723), the one from development works fine.

What about the other builds, i.e. **Windows**, **Linux** and **Android**? Do we also care about Chrome, Blackberry and FlasCC right now?

If all hosts are in equal or better shape on development than on master, I think we should merge (given Patrick's approval).


superquadratic:
Btw, while GitHub says it "can’t automatically merge this pull request", there are no merge conflicts when merging using the command line.

PR #746, opened 2013-08-26
fix text box not displaying some characters

Vavius:
I don't know for sure why that line was there, but it was preventing rendering of some characters. For example, Thai font have a lot of characters (vowels) that should be rendered above consonants and their xAdvance is 0.


CircuitCipher:
Can you provide a way to reproduce the problem so that we can verify that this fix works? Being English speaking it would be helpful to have:
- A link to a font that can reproduce this
- What characters on a US keyboard to type with the above font to reproduce this problem
- What the expected results look like
- What the bad results look like
  
  Thanks.

Vavius:
Here is small repro sample with Thai font included. 
Pay attention to glyphs that appear above another glyphs. 
https://github.com/Vavius/MOAITextBox-repro/tree/master/font-thai

My forum thread with screenshot: http://getmoai.com/forums/post10786.html

CircuitCipher:
Thanks, I'll look at this tonight.

CircuitCipher:
Okay, it looks good.  If you can, please do the following:
- Take your sample code and put it under samples/test/textbox-thai
- Add some comments in the main.lua stating
  - What the test is testing for
  - What the changes were to fix the problem

Once you add those in, I'll merge the changes.  Thanks again!

PR #747, opened 2013-08-27
Merge Thai Text Rendering Changes to Develop.

CircuitCipher:


superquadratic:
@joshua-jandyco,

I noticed that you are using pull requests a little different than I am, so I'd like to share some of my best practices regarding the workflow. Since we don't have a mailing list for maintainers (yet?), I'll just put it here:

I usually don't merge my own pull requests. The idea of a pull request is to put one's code up for discussion and review by the other maintainers. Don't take this the wrong way – your contributions have been of great quality. But every now and then, a second pair of eyes will see a better fix or spot an issue that would otherwise go by unnoticed.

On the other hand, when merging something from master into development (after merging a pull request into master, like in this case), I don't think we need to open a pull request for that. When it's approved for master, it's also approved for development.

Let me know what you think!
// Guido

CircuitCipher:
This particular one I just did with a pull request because I don't have access to a computer at the moment to do the merge of the code in master so it was the quickest way to I could find in the Github UI to accomplish it.

I agree about pull request merging and reviews.  If we have a diligent set of contributors to keep on top of the process and perform the reviews and merges that is fine.  

As long as others review and merge in a timely fashion and we don't end up with the same problem as before with a stagnant list of pull requests, I'm all for it.  I'll leave up future requests for peer review and merging.

superquadratic:
> This particular one I just did with a pull request because I don't have access to a computer to do the merge of the code in naster so it was the quickest way to I could find in the Github UI to accomplish it.

Ah, I see!

> I agree about pull request merging and reviews. If we have a diligent set of contributors to keep on top of the process and perform the reviews and merges that is fine.

Cool. Having at least three active maintainers (you, @makotok, me), I think there's a good chance of getting pull requests merged in a timely manner. :)

CircuitCipher:
Sounds good to me.  Thanks.

makotok:
I want many more maintainers.
There is no choice but to always looking.

Issue #566, opened 2012-09-14
Better parity between getters/setters


patrickmeehan:
As a user I would like better parity between getters/setters so that I do not have to store extra copies of variables in Lua tables.

PR #748, opened 2013-08-28
Add some getListener() functions

superquadratic:
I recently needed these getters, so I implemented them. A small step towards fixing moai/moai-dev#566. :)

### DIFF
--- a/src/moai-core/MOAIEventSource.cpp
+++ b/src/moai-core/MOAIEventSource.cpp
@@ -65,6 +65,25 @@ void MOAIEventSource::SetListener ( lua_State* L, u32 idx ) {
 //================================================================//
 
 //----------------------------------------------------------------//
+/**	@name	getListener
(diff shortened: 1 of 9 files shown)


CircuitCipher:
This looks good.  We should keep making progress towards this as I see a lot of newcomers ask about why getters are missing.

@superquadratic Could you look into adding the same to the Android versions of the extensions you modified, please?  Should basically be the same code.  Thanks.

superquadratic:
`setListener()` is copy-and-pasted in every Android singleton, so we will either have to do the same for `getListener()` (ugly) or refactor both to use `MOAIGlobalEventSource` (risky). Unfortunately I don't have the resources to implement and (more importantly) test either option.

CircuitCipher:
Okay sounds good. We can address that at a later time.

CircuitCipher:
Made a test file to test it out.  Looks good.  Thanks. I'll merge this in.

PR #749, opened 2013-08-28
Update Xcode projects to recommended settings

superquadratic:
Xcode compained about some settings, so I let it update the projects.


CircuitCipher:
Looks like it mainly changes the default debugger to LLDB which is good.  I'll merge this in.

PR #750, opened 2013-08-29
Fix MOAITimer.EVENT_STOP

superquadratic:
The new `MOAINode::EVENT_UPDATE` (enum value 0) collided with `MOAIAction::EVENT_STOP` (also enum value 0) in `MOAITimer`, which is derived from both Node and Action. As a result, `EVENT_STOP` was triggered every frame.

Fixed by starting `MOAIAction` events at `MOAINode::TOTAL_EVENTS`, similar to `MOAITimer` events starting at `MOAIAction::TOTAL_EVENTS`.

### DIFF
--- a/src/moai-sim/MOAIAction.h
+++ b/src/moai-sim/MOAIAction.h
@@ -5,6 +5,7 @@
 #define	MOAIACTION_H
 
 #include <moai-sim/MOAIBlocker.h>
+#include <moai-sim/MOAINode.h>
 
 //================================================================//
 // MOAIAction
@@ -67,7 +68,7 @@ public:
 	DECL_LUA_FACTORY ( MOAIAction )
 	
 	enum {
-		EVENT_STOP,
+		EVENT_STOP = MOAINode::TOTAL_EVENTS,
 		TOTAL_EVENTS,
 	};
 	


CircuitCipher:
Good catch.  This is definitely something to watch out for when creating these class hierarchies that have events.

I did notice that it seems that the new event in MOAINode is not listed in the doc comments.  For little things like this do you think we should add an issue ticket so that we remember to do them?  These type of things might make good first time contributions for newcomers.

CircuitCipher:
Alright, so I will merge this in.  To me it's slightly strange how the event id's are because MOAINode and MOAIAction are not related until they are both base classes of MOAITimer where this issue comes in and so we use MOIANode's TOTAL_EVENT in MOAIAction to start numbering events.  However, hypothetically classes could be base classes in any number of configurations that could cause conflict.  (Not saying it's likely).

It's in scenarios like this that I prefer Java's style of Enum.  You don't run into these issues.  I wonder if our event id's should be string based so they could be namespaced better and not cause conflict.  Are there reasons against this?  (Just thinking out loud).  Also, http://stackoverflow.com/questions/1965249/how-to-write-java-like-enums-in-c (in this case you would use a string to map the Enum into Lua)

superquadratic:
I totally agree, this is strange. Sooner or later, this will bite us again, but I'm not sure how to fix it for good.
- Strings as event IDs would be a solution, but maybe there are performance implications?
- I was also wondering if we could register events at runtime, then we could just use a global counter to guarantee unique integer IDs.

As for the missing documentation, I'll open an issue for that.

CircuitCipher:
I created a new issue for re-evaluating how we are doing event ID's.  See issue #753.

PR #760, opened 2013-08-31
Feature/ios hide keyboard

makotok:
I added a hideKeyboard method in the same way as Android.
With this fix, can be operated the keyboard freely.


CircuitCipher:
Tested and works.  Merging.

PR #761, opened 2013-08-31
Fixed an issue that is not GC by strong reference from the runtime.

makotok:
I have deleted the code that is very problematic.
This code, make a strong reference from the LuaRuntime.

As a result, many objects will not be GC.


patrickmeehan:
Thank you, Makoto. I am sending the same along with some comments here:

https://github.com/moai/moai-dev/pull/769

makotok:
Thank you for fix.
I will confirm that.

PR #766, opened 2013-09-01
Fix setClearColor to accept no arguments

Vavius:


Vavius:
Simple old bug. I remember someone had fixed that. But it's not?

superquadratic:
Since this is a bugfix, I merged it directly into master in 032befdff62863e2f9f8632ed0b6cc39316ce832.

PR #769, opened 2013-09-03
Revisions to Moai Lua binding, memory management and garbage collection

patrickmeehan:
This is just to get the discussion started. I revised the way Moai thinks about Lua GC to handle circular references quite a while ago, but there are still some edge cases that can crop up.

This particular pull requests restores the last 'good' behavior that avoids circular references and allows the GC to operate normally. I mistakenly added a call to MakeStrong in MOAILuaObject::OnRetain (), which fixed a sporadic crash I was seeing but inadvertently reintroduced circular references.

The problems stem from the divergence of the engine's internal ref counting sematic and Lua's GC. To avoid circular references, the engine's Release/Retain is decoupled from the Lua GC - Lua can collect an object but the engine can hang on to it. Likewise, the engine can be done with an object but Lua will hang on to it. 

I would rely entirely on the Lua runtime entirely for managing references between objects, but performance is a concern. There are cases where the engine will want to ensure an object exists (via Retain) and not have to go out to the Lua runtime every time it needs access.

I think there are some intermittent crashes where an object gets GC'd but is kept around by the engine which then expects the object's Lua-side references to still be valid. For example, the object might have a callback that the engine needs to invoke, but if the callback was also collected then its ref will be stale. The ref will _appear_ valid to the engine and (worse) may have already been re-purposed to refer to something else leading to sporadic and hard to debug crashes.

We can keep the current system, but the every use of MOAILuaRef, MOAILuaLocal and Release/Retain needs to be audited. It is also not safe to take the bool value of a MOAILuaLocal or a weak MOAILuaRef at face value as these objects may have been garbage collected and the ref number reused. Probably weak MOAILuaRef's should be avoided altogether and MOAILuaLocal's should never be used if their owner has been flagged for garbage collection. 

I'm pushing this fix (and comments) mainly for the discussion. Will be doing more work on this in the coming days.

### DIFF
--- a/src/moai-core/MOAIEventSource.cpp
+++ b/src/moai-core/MOAIEventSource.cpp
@@ -76,7 +76,7 @@ int MOAIInstanceEventSource::_getListener ( lua_State* L ) {
 	MOAI_LUA_SETUP ( MOAIInstanceEventSource, "UN" );
 
 	u32 eventID = state.GetValue < u32 >( 2, 0 );
-	if ( !self->PushListener( eventID, state )) {
+	if ( !self->PushListener ( eventID, state )) {
 		state.Push ();
 	}
 
--- a/src/moai-core/MOAILuaObject.cpp
+++ b/src/moai-core/MOAILuaObject.cpp
@@ -24,13 +24,17 @@ MOAILuaLocal::~MOAILuaLocal () {
 //----------------------------------------------------------------//
 int MOAILuaObject::_gc ( lua_State* L ) {
 
+	// This gets called when the object falls out of Lua's scope
+	// The object may still be referenced by the engine, in
+	// which case it isn't ready to be deleted
+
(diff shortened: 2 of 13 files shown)


makotok:
I Fixed a compile error of some.
Confirmed the operation.

For now, GC is working well.
Behavior of MOAITask is unconfirmed.

patrickmeehan:
Thanks, Makoto!

superquadratic:
I don't want to end the discussion here, I just merged it since it "restores the last 'good' behavior" (as Patrick said), so I think it's a good starting point for the changes to come.

Makoto, I applied your fix in 325ecbeb024e4e0fa35740f5531043eec7f5cf3d.

patrickmeehan:
Thanks! Got busy again but verifying the mem management approach and writing some Lua torture tests for it is otherwise top of mind. Soon(tm).

Issue #594, opened 2012-10-24
MOAIPathFinder does not use a*


jrabek:
For the same computational/memory overhead, MOAIPathFinder can be using a\* and finding much better paths than the current implementation.

The current implementation only uses the heuristic to the target and does not add in the cost of the path traveled so far resulting in circuitous paths in some cases.

I have implemented a\* on the following branch:
https://github.com/jrabek/moai-dev/tree/a-star

I wasn't sure if this was intentional so I am filing this as a bug instead of a pull-request.

Let me know if I should just open a pull request instead.

Also, in the same branch I pushed a separate commit that will path find for rectangular entities larger than one grid square (bosses, groups, etc)

jrabek:
https://github.com/jrabek/moai-dev/commit/9197ad6b2a17d57e986bed7582ba240e5493c7fb

adamvillalobos:
Thanks, we'll take a look at this.

DanielSWolf:
I believe this issue can be closed. @superquadratic implemented A\* pathfinding in issue #770, which has been merged.

makotok:
Thank you for information. Close.

PR #770, opened 2013-09-04
Support Path Finding in Arbitrary Graphs

superquadratic:
This adds support for path finding in an arbitrary (i.e. non-grid) graph. This graph is represented by `MOAIVecPathGraph` and can be used with `MOAIPathFinder`. A sample demonstrates the new functionality.

I had to tweak the algorithm in `MOAIPathFinder` a little to make sure it finds the shortest path, not just any path. The change is to take the cost of the path so far into account when finding the best path to pursue. If I'm not mistaken, this is now a complete implementation of A*. See also moai/moai-dev#594, which can be closed when this gets merged.

I have updated the project files for OS X, iOS and NaCl. Android and CMake seem to use globbing and should work out of the box. Could anyone add MOAIVecPathGraph.cpp/.h to the Visual Studio projects after merging?


makotok:
Add to Visual Studio ask someone.

PR #773, opened 2013-09-07
Overhaul of Grid/GridDeck2D/Prop interactions

seebs:
It would be neat if grids had the ability to set color and scale
for each object. But it would also be neat if this weren't really
expensive.  So! We create a new DrawRegion function for the grid,
which just calls another function through a function pointer. The
existing implementations of this logic in GridDeck2D and Prop are
combined into a function, with a little generalization. Then four
copies are made, one for each combination of color being present
(yes/no) and scale being present (yes/no), so if you don't add
colors and scales, you don't get a ton of extra processing. ZLColorVec
has an operator\* now.

The actual implementation of the color is a bit ugly; we find the
current "ambient" color of the GfxDevice once at the start of the
loop, then for each grid element set the ambient color to the
starting color times that grid element's color, then restore the
original value after the loop. The reason for this is that the
alternative is to introduce another overload of Draw() which accepts
a color argument, and use that for _everything_, which is much
more intrusive. (On the other hand, it might be more efficient.
On the other, other, hand the way that polygons are getting rendered
right now is already running through multiple layers of indirection,
so I'm not sure it matters.)

Signed-off-by: Seebs seebs@seebs.net


makotok:
Do you have a sample to confirm this fix?
I want to add the sample.

Also, please request to development branch new features.
The fix appears to be some risk.

https://github.com/moai/moai-dev/blob/master/CONTRIBUTING.md

seebs:
I haev updated the commit to include a sample (merged into the previous commit). I don't have a clear idea of how to change which branch it is requested against, though -- github's interface confuses me.

CircuitCipher:
I'm not sure of an easy way to switch which branch this pull request is against.  You may have to close it and reopen a new one.  You will have to be sure that the branch you worked on is based on development.

CircuitCipher:
Closing this in favor of the one against development branch, #777.

PR #774, opened 2013-09-08
Fix to allow LUAEXT to work on linux.  

jacobgardner:
I was unable to use socket and various other lua extensions without these changes.  If anything should be changed, please let me know and I'll fix it within a day or so.  

In the build-linux_x64 script, MOAI_LUAEXT needs to be set to true and MOAI_OPENSSL needs to be set to true.  They are both false in this commit as that was the state of the script before my modifications.

Once changed to true,
./bin/build-linux_x64 works as expected.

Reasoning behind changes:
I had to split some of the third-party libraries (crypto and curl) into their own projects due to an issue during the linker phase.  crypto and luasocket, both have a buffer.c.  Because they were both being compiled in the same project (third-party) they were creating two object files with the same name (buffer.c.o).  When third-party.a was being linked together the second buffer.c.o to be linked overwrote the first and as a result the functions defined in the first could not be found.  Likewise select.c in curl was conflicting with the one in luasocket.  This also needed to become it's own project to resolve this issue.


makotok:
Cmake does not work on OSX.
I are working on fixing.

CircuitCipher:
@makotok Thanks, I look forward to your solution.  I did a little bit of work on that but decided to put more time into the LuaJIT stuff.  I will test your solution when it is finished.

PR #775, opened 2013-09-09
Add LuaJIT support on iOS and OS X

CircuitCipher:
-  LuaJIT build process is automated
  - The new shell scripts are added to the libmoai-ios and libmoai-osx targets as a script build phase
-  LuaJIT can be disabled by following the instructions in the README
  - Remove linking of LuaJIT
  - Set the MOAI_WITH_LUAJIT variable to 0 
  - Link against the newly created targets libmoai-osx-lua and libmoai-ios-lua

I'm aware of other efforts such as issue #662 which was closed by Patrick stating that they wanted to wait for cmake build system.  

Currently Cmake is broken for OSX so I feel this PR will at least allow users of Moai to make use of the performance that LuaJIT offers until this problem is resolved.

### DIFF
--- a/xcode/README
+++ b/xcode/README
@@ -1,3 +1,43 @@
-To build any of the FMOD targets in Xcode you will need to add FMOD_IOS_SDK_ROOT and FMOD_OSX_SDK_ROOT using Xcode's 'Source Trees' feature.
+Building Without LuaJIT
+==============
 
-To do this, from the Xcode menu go to Xcode/Preferences/Locations. Choose the 'Source Trees' tab. Add FMOD_IOS_SDK_ROOT and FMOD_OSX_SDK_ROOT with the corresponding paths to the *root* of the respective FMOD SDK.
+Unset Xcode LuaJIT Environment Variable
+--------------
+* Open the libmoai Xcode project
+* From the left bar, show the project navigator (The folder icon)
+* Click on the top level project (It has the Xcode project icon)
+    * In the main window you will see all the targets for the project
+* Select libmoai under the "Project" header
+* Click the "Build Settings" tab
+* Scroll down to the bottom where the "User-Defined" section is
+* Change the value of MOAI_WITH_LUAJIT to 0
+
+Add Lua Target to Your Project
+--------------
+* Open your project's Xcode file
+* From the left bar, show the project navigator (The folder icon)
+* Click on the top level project (It has the Xcode project icon)
+    * In the main window you will see all the targets for the project
+* Click the target for building your application
+* Click the "Build Phases" tab
+
+### Target Dependencies
+* Expand the Target Dependencies section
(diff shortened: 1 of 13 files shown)


makotok:
Build errors occur in Xcode.

```
ld: warning: directory not found for option '-L/Applications/moai/xcode/libmoai/luaJIT/osx'
ld: library not found for -lluajit2
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

In addition, libluajit2.a is displayed on a strange place.
Would you like to add the following target to libmoai.xcodeproj?
- libmoai-osx-luajit
- libmoai-ios-luajit

Also, do you like to be compiled by switching scheme?
- moai
- moai-jit

CircuitCipher:
@makotok  I will fix the linking error tonight.  It looks like I just need to make the link path for luajit to be relative the the group (Which I thought it was already)

## Adding Targets

I didn't add the luajit as a separate target because the target would not produce the library directly.  Please see http://luajit.org/install.html where they specifically say not to directly integrate LuaJIT sources into your code but to use their build script.

I used their build script in a pre-build phase in the libmoai-ios target and the libmoai-osx target.  If you know of a way to create a new target that produces the library file using the official LuaJIT build system then please let me know and I will integrate that into the pull request.

## Scheme Switching

Is there some way to include different target dependencies for each scheme?  Otherwise we wouldn't be able to use that method.  I currently have a user variable that you can set to switch between JIT and Lua builds.

[EDIT]: Alright, I see how we can do separate schemes here: https://developer.apple.com/library/ios/recipes/xcode_help-scheme_editor/Articles/SchemeBuild.html#//apple_ref/doc/uid/TP40010402-CH2-SW1

We would just need to figure out how to create the LuaJIT targets so that the it knows where the LuaJIT build system puts the libs.

superquadratic:
> If you know of a way to create a new target that produces the library file using the official LuaJIT build system then please let me know and I will integrate that into the pull request.

I'm not sure if that helps, but you can add a target with an external build system (Add Target > OS X / Other > External Build System). Using `/bin/bash` as your "Build Tool", you can make it call a shell script that you add to "Arguments". That way, the call to the build script wouldn't be hidden in the pre-build phase.

CircuitCipher:
Thanks @superquadratic that's probably the route I will be going I just have to figure out how to tie the external build output library to the target.

CircuitCipher:
I have incorporated the suggestions given by @makotok.  
- Add libmoai-ios-luajit and libmoai-osx-luajit targets to libmoai
- Add moai-luajit targets to osx and ios projects
- Add moai-luajit schemes to ios and osx projects

[Edit] I forgot, I will have to update the readme as well.  I will do that tomorrow.

makotok:
@joshua-jandyco
Thank you for a nice fix.

makotok:
Product name is duplicated.
I will fix this directly.

You are finished.

PR #776, opened 2013-09-09
Fix/lua gc

patrickmeehan:
I cleaned up the smart pointer semantics a bit and renamed some things to clarify intent. 'MOAILuaLocal' is now 'MOAILuaMemberRef.' I also got rid of MOAILuaObject's GetLocal, SetLocal, PushLocal in favor of the smart pointer. Now all of the Lua smart pointers have a similar interface and usage.

Also also removed the MOAILuaRefTable implementation. Not sure why it was needed in the first place. Now the ref two tables are just a weak and strong table help in Lua's registry and managed with luaL_ref and luaL_unref.

Most importantly, I fixed a crash that would happen when an object's Lua binding would get GC's but member refs remain. The actual crash was an assert () on mMemberTable triggered by a false positive when testing the validity of a member ref and attempting to push the member table.

I've run our game with this and some of the samples, but haven't tested exhaustively. Would definitely appreciate extra eyes on this, particularly the move from MOAILuaRefTable  to luaL_ref  and luaL_unref.


CircuitCipher:
Thanks for this fix!  I just ran into that GC issue you mentioned and am very happy to see you fix it so quickly.
I haven't been able to do a full review yet as I have to head out for the night but so far I found the following issues:

The following classes still reference MOAILuaRef and cause errors due to it being abstract:
- MOAITwitterIOS 
- MOAIAppIOS
- MOAIBillingIOS
- MOAIDialogIOS
- MOAIGameCenterIOS
- MOAIMoviePlayerIOS
- MOAINotificationsIOS
- MOAIPlayhavenIOS

Changing to MOAILuaStrongRef in the headers and changing the SetStrongRef to SetRef in the implementations fixes build issues on IOS.

I haven't had a chance to test Android builds yet.

CircuitCipher:
Android is broken too.  Look for the same type of thing.

patrickmeehan:
OK, cool. I may be back on Android later this week if nobody else has time to fix it. :)

CircuitCipher:
I'd fix both platforms, it seems to be a simple fix.  I'm just not sure how the workflow works in this case.  Do we merge your code in and then do another PR?  In the old workflow where there could have been a moai branch, I could have added directly to this PR.

superquadratic:
> In the old workflow where there could have been a moai branch, I could have added directly to this PR.

That is indeed a shortcoming of the current workflow. One solution would be to pull the changes into a branch in the moai repo (see steps 1+2 of "You can also merge branches on the command line." in the green box below), add more commits to that branch and open a new PR.

CircuitCipher:
Thanks @superquadratic.  I'll do that tonight.  However, after I fix both platforms and have a chance to review the rest of the code, I may just merge the pull request with my changes added via the command line merge.

CircuitCipher:
Alright so before your latest commit, I was still running into crashes.  After the commit I haven't run into any.

However I ran across an interesting and intermittent problem.  As I was stress testing things moving around through the scenes in one of my apps, randomly MOAIRenderMgr.setRenderTable method became nil and I could no longer call it, which stopped my app from functioning.  I was only able to get this to happen twice.  So I'm not sure if this is quite ready yet. I need to play around more to see if I can determine a cause.

Side note, I have the android and ios fixes ready as a patch.

CircuitCipher:
Okay I was able to induce a crash:

```
Thread 1, Queue : com.apple.main-thread
#0  0x00625a14 in luaH_getnum at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/ltable.c:437
#1  0x005ff561 in lua_rawgeti at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/lapi.c:573
#2  0x006023d2 in luaL_ref at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/lauxlib.c:488
#3  0x002672c1 in MOAILuaMemberRef::SetRef(MOAILuaObject&, MOAILuaState&, int) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-core/MOAILuaRef.cpp:208
#4  0x00263a95 in MOAILuaObject::LuaRetain(MOAILuaObject*) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-core/MOAILuaObject.cpp:296
#5  0x0010647e in MOAIAction::Attach(MOAIAction*) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAIAction.cpp:218
#6  0x001063ab in MOAIAction::_addChild(lua_State*) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAIAction.cpp:26
#7  0x0060d68b in luaD_precall at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/ldo.c:319
#8  0x0062bf8a in luaV_execute at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/lvm.c:587
#9  0x0060e20d in resume at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/ldo.c:404
#10 0x0060ce0d in luaD_rawrunprotected at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/ldo.c:116
#11 0x0060dfe0 in lua_resume at /Applications/Moai/moai-official/xcode/libmoai/../../3rdparty/lua-5.1.3/src/ldo.c:427
#12 0x0012e3ab in MOAICoroutine::OnUpdate(float) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAICoroutine.cpp:144
#13 0x00108089 in MOAIAction::Update(float, unsigned int, bool) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAIAction.cpp:420
#14 0x00108285 in MOAIAction::Update(float, unsigned int, bool) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAIAction.cpp:459
#15 0x00109b19 in MOAIActionMgr::Update(float) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAIActionMgr.cpp:142
#16 0x001dd4bc in MOAISim::StepSim(double, unsigned int) at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAISim.cpp:863
#17 0x001dd877 in MOAISim::Update() at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/MOAISim.cpp:954
#18 0x000df743 in AKUUpdate() at /Applications/Moai/moai-official/xcode/libmoai/../../src/moai-sim/host.cpp:398
#19 0x00089092 in -[MoaiView onUpdateAnim] at /Applications/Moai/moai-official/xcode/ios/Classes/MoaiView.mm:273
#20 0x027672d2 in CA::Display::DisplayLink::dispatch(unsigned long long, unsigned long long) ()
#21 0x0276775f in CA::Display::TimerDisplayLink::callback(__CFRunLoopTimer*, void*) ()
#22 0x0384b376 in __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ ()
#23 0x0384ae06 in __CFRunLoopDoTimer ()
#24 0x03832a82 in __CFRunLoopRun ()
#25 0x03831f44 in CFRunLoopRunSpecific ()
#26 0x03831e1b in CFRunLoopRunInMode ()
#27 0x036d57e3 in GSEventRunModal ()
#28 0x036d5668 in GSEventRun ()
#29 0x028c9ffc in UIApplicationMain ()
#30 0x00089ef5 in main at /Applications/Moai/moai-official/xcode/ios/main.mm:12
```

patrickmeehan:
Hmmm. Think you can put together a repro script I can test against?

CircuitCipher:
I'll see if I can find a way to reproduce it and distill it down tonight.

CircuitCipher:
As I said, I ran into these 2 issues just by jumping around the scenes in one of my apps.  I have yet to come up with a consistent way of reproducing the issues.  I do notice I see these messages once and a while usually before the crashes:

"Attempt to access MOAILuaObject userdata tagged for garbage collection; running a full cycle of GC prior to reallocation."

@makotok and @superquadratic Have you guys had a chance to try these changes out?  If need be I can post my patch for ios and android so you can easily test them.

makotok:
@joshua-jandyco 

```
I'd fix both platforms, it seems to be a simple fix. I'm just not sure how the workflow works in this case.
Do we merge your code in and then do another PR? In the old workflow where there could have been
a moai branch, I could have added directly to this PR.
```

I agree.
Create a direct branch if necessary.
Fixed in everyone.

superquadratic:
I ran a quick test with our game (on OS X) and got an intermittent [crash](http://pastebin.com/6hMiJKeV) when quitting the Glut host. Haven't had the time to look at the changes in detail, though.

patrickmeehan:
@superquadratic I've been seeing that too. Looks like the host is trying to call the update after the context has been destroyed. Not sure why this is happening now - haven't touched the host.

Tracking down these crashes is going to be my weekend.

patrickmeehan:
@joshua-jandyco DId you see that crash after "Attempt to access MOAILuaObject userdata tagged for garbage collection; running a full cycle of GC prior to reallocation." just with this branch, or has that happened before? I reverted to using Lua's ref system for the weak and strong ref tables - am wondering if that has something to do with it. The MOAIObject userdata is held in a weak table, so may be tagged for garbage collection.

CircuitCipher:
@patrickmeehan I can't remember for certain but it was either only in this branch or with the previous GC related changes that were merged into development that I started seeing that message and the crashes.

Before those I know that I have never seen that message.  If you would like I can get you a definitive answer but I probably won't be able to do the tests until tomorrow.

patrickmeehan:
No rush. The rub with this is that I don't have an easy way to test the case where the gc tags a userdata for collection and Moai tries to access that userdata before the __gc metamethod is called. This problem might have existed all along but since I was using a bespoke reference table in the previous version we didn't see an assert ().

My thought now is to detect the case, explicitly unbind the userdata that's been tagged, set its __gc to nil then re-bind the MOAILuaObject.

The cases are:

1 - Moai deletes an object that hasn't been gc'd. We should clear the userdata's __gc, attach a 'tombstone' to it, release any refs to it and leave it to Lua to clean up.

2 - Lua gc's an object and the__gc metamethod is called. This is the happy path. Release the refs and delete the object if it's no longer needed.

3 - Lua tags a userdata for GC but doesn't call __gc. Moai tries to access the userdata and gets nil. We should clear out the __gc and release the refs. No need to attach a tombstone as the userdata has gone out of scope. Depending on why the engine was trying to access the userdata we may need to create a new userdata and re-bind.

What still needs to be worked out:

4 - What happens to the member refs that use the object's member table when the object gets re-bound to Lua? Those need to be cleared out.

5 - What happens to objects being held on to via LuaRetain? They may have also been gc'd or tagged for gc. If so they need to be re-bound along with the retaining object.

I also wonder if there's a simpler approach. It seems like the case where the engine needs an object after it has gone out of Lua's scope is the root of all evil. The two approaches I can think of are:

6 - Get rid of engine ref counting altogether. This means that the engine sim systems (action manager and node graph) will have to add and remove objects from a global Lua table. In the case of the node graph, this could get expensive as every node scheduled for updated would need to be added then removed on every frame. Yuck! The alternative is to add explicit memory management commands to nodes and put the burden of adding/removing them on the script. Double yuck!

7 - Keep something like the status quo but let the engine manage Lua's gc so that it will never run during an action tree or node graph update. The problem there is I don't know if that would really solve anything - we still need an engine-side memory manager unless we require every object to be bound to Lua. In that case, we're back to the engine systems having to add and remove objects from their own tables every frame. Not so bad for the action tree, but probably a deal breaker for the node graph.

Another option, I suppose, is to only allow an object to get bound to Lua once. So an object will (typically) never go out of Lua scope then get handed back from the engine. I think that would let me get rid of the annoying 'LuaRetain' idea which aims to keep a Lua binding around based on inter-object relationships. In cases where the engine creates a default object (like the root action) the default can be handed to Lua once, but then if lost it's lost forever. So something like:

local root = MOAIActionMgr.getRoot () -- default root is bound and returned
root = nil -- root goes out of scope
-- gc runs here...
root = MOAIActionMgr.getRoot () -- return nil and/or report an error and/or throw an exception

This would also make expiring 'member' refs a little easier. If we allow re-binding then MOAILuaMemberRef needs either a binding ID that's guaranteed to be unique (tricky) or to be kept in a singly linked list and released on unbind.

patrickmeehan:
OK, what do you guys think of that final option to simplify things? To recap:
- Only allow engine objects to be bound to Lua _once_.
- Get rid of 'LuaRetain' - object A will not cause object B to preserve its Lua binding if object B goes out of scope.

The tradeoff is we force an error or unexpectedly return nil in this case:

local root = MOAIActionMgr.getRoot () -- default root is bound and returned
root = nil -- root goes out of scope
-- gc runs here...
root = MOAIActionMgr.getRoot () -- return nil and/or report an error and/or throw an exception

Actually, we'd probably just remove all object getters and force scripts to track them explicitly so the above case can never arise.

bullno1:
There's a problem with binding once: prop picking. I think the userdata of props can be gc'd if you have many props in a scene. However, they should be retrievable using picking methods (propForPoint ...). Returning nil on a successful pick is troublesome. We can return lightuserdata though. All objects can have a getId method which returns a lightuserdata with the value of its C++ object address. It's still pretty weird.

But if we use a table at engine side to keep them alive. We'll run into the same problem again.

Most users have some high level framework anyway, so that might not be an issue. However, tutorials must be written with global variables to prevent surprises.

tommo:
I haven't read much about MOAI memory management yet, please forgive me if I'm saying something wrong.
It seems that object deletion can be done in both smartpointer and __gc.
Why not let __gc only reduce the refcount,  and leave finalization job to smartpointer ?

The workflow will be:
- Bind Lua Object  ->  refcount +1  ,push a weakref in lua registry
- Push object to Lua -> if weakref alive,use it.  if dead, rebind.
- Lua Ref collected  -> refcount -1

In this way, Lua reference always holds 1 refcount as long as it's alive.

Lua handles cycle reference, you can forget about Lua side and focus on C++ side.

bullno1:
It's exactly what's happening now. The problem is Lua knows nothing about an object's internals so it can't handle cyclic reference at all.

bullno1:
How about this:
All Lua-bound objects have a table to track its references.
There's a global table with weak keys. It maps an userdata to its reference table. Thus, as long as an object is gc'd, all references are gone. 
Root objects (render table, action root...) are kept alive using lua_ref. 

Effectively, we built a reference graph which is accessible to Lua GC. 

This way, we can still expect old behaviors (object relations keep userdata alive).

patrickmeehan:
That's the current model (if I understand correctly). Each object has a local table that tracks its own references. Check out MOAILuaMemberRef. The problem is when an object is tagged for gc or gc'd then needs to be recreated. All of its member refs then also have to be recreated. So to make everything work OK I'd need to add a singly linked list to each object of all its member refs so I can traverse them.

Maybe not the end of the world to re-bind on demand.

bullno1:
But why would an userdata get gc'd if it's still being used? And how do you keep that local table alive?

I searched through the code and the only instance of __mode is in MOAILuaRef. Both key and value are weak while the table has a luaL_ref pointing to it.

Let me rephrase my suggestion:

Each object has 2 parts: userdata (pointer), member table to track references.
There's a global table in the registry with __mode='k' called "memberTables"

When an object is bind to Lua for the first time:
- Create userdata: udata
- Create member table: memberRefs
- Track member table: memberTables[udata] = memberRefs

With only this, everything will be gc'd since they are all circular references. Thus we need a root set. Only objects like (current) render table or  (current) action root are added to this root set using luaL_ref.

The root set is of course, managed by C++.

Userdata will stay alive as long as C++ or Lua needs it and a whole subgraph can be gc'd when it's unreachable from both sides.
Following this, MOAIPartition or its family (cell, level) would also need to track references.

patrickmeehan:
I think that's what we have now, the difference being that the member table is attached to the userdata directly. It's in the per-object stack that gets built when the object is bound, so the intent/behavior is the same.

My concern about tracking every object relation in a member table is performance, particularly around the node graph. It's fine to have MOAIPartition track all its members because props usually don't get added/removed with high frequency, but adding/removing graph nodes during the dep graph update could really hurt performance.

That said, I like your idea of moving the member table out of the object's userdata stack, but modified to use the object's address to track the member table (so it can be decoupled from the userdata). That would handle the case where we need to create an object ref during construction, before the userdata exists (this may be the source of the intermittent crash during action tree update - the root action isn't being properly pinned by the action manager). 

We'd then wind up with memory management semantics that look like something this:

MOAILuaStrongRef - use a strong registry table to track Lua refs
MOAILuaWeakRef - use a weak registry table to track Lua refs
MOAILuaMemberRef - use the member table to track Lua refs
MOAILuaObjectRef - (formerly MOAILuaSharedPtr) use the member table to track Lua refs as well as engine objects

Global managers just take a strong ref back to their member table.

Another implication of "userdata will stay alive as long as C++ or Lua needs it and a whole subgraph can be gc'd when it's unreachable from both sides" is that we should never need to re-bind or access a member or object ref once its owner has been gc'd. I think this also lets us remove the ref count from the MOAILuaObjectRef implementation (and LuaRetain/LuaRelease). The engine ref count would then be strictly for temporary use during updates and Lua would handle all other memory management. We could do away with MOAISharedPtr and MOAIWeakPtr entirely and just use Release/Retain in a very few places. Could almost even go as far as moving the ref count out of MOAIObject and treat it as a mix-in just for MOAINode.

patrickmeehan:
Where we're headed seems to be: use Lua to do all long-term memory management. In a few cases where performance is important and object membership in an engine manager class can be thrashed while the gc runs (the dep graph update, in other words) use reference counting but with the understanding that once an object has been gc'd by Lua its destruction is imminent.

I think this will simplify things a lot. Unless somebody objects I'm am going to put together an implementation on this branch.

[edit]

Could alternatively do away with ref counting altogether by allowing the engine to turn off Lua's gc during critical operations such as dep graph update. Would then need to replace the Lua gc control function with Moai equivalents so Moai can be aware of what gc params to restore.

[/edit]

patrickmeehan:
Another tangentially related question to chew on is whether attribute links should also create inter-object Lua references. In the current implementation they do _not_. The reason being is that timer-based dep nodes are also actions, which get pinned by their parents, and the rest tend to be props which get pinned by their partitions.

This leaves script nodes and pure transforms which do not and may be gc'd away.

Who here thinks its worth the overhead to make/break Lua refs for attribute links? Also, who thinks these links should be symmetrical vs. one-way (i.e. only on push or only on pull)?

bullno1:
After reading your comment and MOAILuaObject.cpp , I don't understand why a userdata can be gc'd while the engine is still using it. Is it due to bad reference tracking? Also, is there a case where C++ creates cyclic reference? If there isn't, isn't it enough to let a C++ reference pin the underlying userdata? That's what LuaRetain is doing right? 

About gc, I think it's ok to turn it off during critical operations. It may be even better to turn automatic gc off entirely and control it:
http://www.altdevblogaday.com/2011/06/23/from-python-to-lua/
MOAISim knows about the frame time budget in a fixed frame rate setting so it should be put in control of the GC.

patrickmeehan:
Yeah, it should be fine for a userdata to get gc'd while the engine is using its object, but in practice there seem to be some edge cases. What you're describing is the current approach - let Lua kill the userdata, pin objects we need from C++ and take care not to create any circular references. It's just a little ugly to tear down then rebuild the lua binding... and trying to pin the Lua binding is feasible but also sort of ugly. So I'm really warming up to the idea of letting Lua be the final word in memory management.

Thanks for your thoughts on turning off the gc (and the link). I tend to agree.

bullno1:
A weird stretch case for attribute: Let's say prop A pulls from pure transform B and pure transform B pulls from prop C. Both prop A and C are pinned by their partitions. What will happen if B gets gc'd? Should the A -> C link suddenly stops functioning?

To "add an extra layer of indirection" to solve a problem, this situation may happen. Having attribute links track objects sounds safer.

makotok:
A fix for this issue or would be required?

Disadvantage for the fix is because large.
In addition, I can work around the problem easily.

If the object is required, to have in themselves.

1) NG Pattern

``` Lua
local node = MOAIScriptNode.new ()
node:setCallback ( onUpdate )
node:scheduleUpdate ()
node = nil
collectgarbage ( 'collect' )
```

2) OK Pattern

``` Lua
local node = MOAIScriptNode.new ()
node:setCallback ( onUpdate )
node:scheduleUpdate ()
coroutine.yield()
node = nil
collectgarbage ( 'collect' )
```

3) OK Pattern

``` Lua
local node = MOAIScriptNode.new ()
node:setCallback ( onUpdate )
node:scheduleUpdate ()
node = nil
MOAISim.forceGarbageCollection()
```

superquadratic:
Closing, since these change sets (and a lot more work on memory management) is included in #787.

PR #777, opened 2013-09-10
Overhaul of Grid/GridDeck2D/Prop interactions

seebs:
It would be neat if grids had the ability to set color and scale
for each object. But it would also be neat if this weren't really
expensive.  So! We create a new DrawRegion function for the grid,
which just calls another function through a function pointer. The
existing implementations of this logic in GridDeck2D and Prop are
combined into a function, with a little generalization. Then four
copies are made, one for each combination of color being present
(yes/no) and scale being present (yes/no), so if you don't add
colors and scales, you don't get a ton of extra processing. ZLColorVec
has an operator\* now.

The actual implementation of the color is a bit ugly; we find the
current "ambient" color of the GfxDevice once at the start of the
loop, then for each grid element set the ambient color to the
starting color times that grid element's color, then restore the
original value after the loop. The reason for this is that the
alternative is to introduce another overload of Draw() which accepts
a color argument, and use that for _everything_, which is much
more intrusive. (On the other hand, it might be more efficient.
On the other, other, hand the way that polygons are getting rendered
right now is already running through multiple layers of indirection,
so I'm not sure it matters.)

Added: New sample, "grid/grid-color-scale", illustrating how the
feature works.

Signed-off-by: Seebs seebs@seebs.net


patrickmeehan:
Please give me a little more time to look at this and make some suggestions. In general I think it's great but there are a few things I'd like to review.

I'm assuming anybody chomping at the bit to use this very cool contribution can pull directly from seebs? Just let me know if anyone is blocked on this and I'll try to prioritize it.

In the interim, maybe subclass MOAIProp, MOAIGrid and MOAIGridDeck and offer these changes in an extension as a 'fancy' version of those?

seebs:
I thought about subclassing, but I am not super good at C++ and didn't want to. The biggest risks I saw were the memory and runtime computation, which is why the grid makes allocation of the scale/color things optional, and uses that fancy function pointer stuff to avoid doing any computations when it doesn't need to. Could probably be done more cleanly.

The other thing that's since occurred to me that might be worth it for large grids would be a palette-based option, where you set color as an integer in a small range, with an associated table of color values...

CircuitCipher:
I agree with @patrickmeehan that a subclass may be a better way of handling this.

### Draw Methods Refactor

I brought this up to @seebs in IRC and would like to see what others think.  Currently there are multiple draw functions for the scenarios where there are color and scale, just color and just scale.  

My idea was to have the methods where scale/color is applied (Like AddColor SetScale) just add a function pointer to an array of functions to be applied in the draw call.  The draw method would loop over each function and apply the necessary operations. This makes the code a little more DRY.  However, performance would have to be a consideration in either case.

I can expand on this more if someone needs me to.

seebs:
It may be premature optimization, but basically I wanted the actual grid drawing loop to be direct calls whenever possible so it could be inlined and optimized decently. My goal was to minimize the number of times we have to make a call through a function pointer which necessarily precludes inlining.

CircuitCipher:
Thank you for expanding on the performance concerns @seebs.

patrickmeehan:
I really want MOAIGrid to stay conceptually clean as a grid of indices laid out in space. May even want to go as far as separating those notions again. (In an earlier version of the framework, the data and spatial components of MOAIGrid were separate: a list of indices is a different thing than a spatial layout.) The grid should be just that: a grid. Would like drawing and styling to be bound at the prop level through layers of additional data provided in decks.

Would like to move toward an architecture where MOAIProp aggregates a series of styles over an index or set of indices that may be laid out in a variety of ways - how the indices are addressed (one or two dimensions) is orthogonal to how they are laid out. So we'd wind up with three different ideas that can be mixed and matched to create a variety of effects. For example, right now there is one deck per prop, but you could imagine a scenario in which multiple decks may be attached and their contents combined via mixing semantics. That way you could use an array of color vecs or transforms to tint or offset any kind of deck being used in any kind of layout.

seebs:
Hmm. We already have flags embedded in the grid, which sort of violates that. But I do see the benefit here. It might be nice to modify MOAIGrid into a template MOAIGrid<t>, with the existing usages turning into MOAIGrid<int>. Then a Prop could in theory have SetTileGrid (aka the current SetGrid), but it could also have SetColorGrid/SetScaleGrid, and those would use a Grid<ZLColorVec> or Grid<float>. But that is sort of a breaking change, so that might want to go into a new class.

At which point my lack of C++ clue becomes a problem: How hard would it be to make a hypothetical template class MOAINewGrid<t>, and then just tell the compiler that MOAIGrid => MOAINewGrid<int>?

Anyway, I'll see if I can come up with a prototype using this idea, or something like it, because I do think this makes more sense. Maybe we should move the discussion to the forum, and close out the pull request for now, since it seems like this design doesn't match your design goals for Grid.

patrickmeehan:
@seebs By all means sketch out an implementation if you'd like or just leave what you have. I will pull your branch into my own repo either way and use what you've done as a starting point - you've already worked out a lot of the performance issues and tradeoffs. And I agree about the flags. It did occur to me to add a size option to MOAIGrid instead and to let whatever does the styling be configured to use a smaller grid (8 or 4 bit, for example) as the flags. I think the answer will become evident when I get into it.

superquadratic:
@seebs: Is this PR still valid, or is it superseeded by #789?

seebs:
Yes, 789 is better.

PR #780, opened 2013-09-14
Latest changes from 'studio' branch

patrickmeehan:
I've only built this for vs2008, Lua 5.1 Debug. Maybe some fixes needed for the other platforms.

After this has been merged and built we should drop the studio branch and never speak of it again.  :]


CircuitCipher:
It's hard for me to tell with the online interface, but I can't determine if the Android Play services additions are merged or will merge in with the Twitter functionality appropriately.  I will have to look at this on my machine and verify.

makotok:
I will confirm these corrections.
First, starting from incorporating this fix.

makotok:
I was built by merging.
However, it is an error in the source below.

```
/Users/mkinoshita/moai-dev/src/moai-sim/MOAICoroutine.cpp:5:10: 'aku/AKU.h' file not found
```

``` C
// Path is old.
#include <aku/AKU.h>
#include <moaicore/MOAIActionMgr.h>
#include <moaicore/MOAILogMessages.h>
#include <moaicore/MOAICoroutine.h>
```

I can not merge this fix.
Better to create a new branch and commit only necessary modifications would be fast.

superquadratic:
Closing, since these change sets are also incuded in #787.

PR #782, opened 2013-09-17
Seebs/grid dev2

seebs:
Okay, this is sort of partial at best, but this is about as much as I know enough to do.

This is a rework of the previous grid design, but this time using separate grids of different types. So, the base MOAIGrid type behaves the same way it used to, but we now also have a MOAIFloatGrid and MOAIColorGrid available, which use float and ZLColorVec respectively. MOAIProp gets new SetColorGrid and SetScaleGrid (and Get...) members, and the sample is updated.

To merge this usefully, you would also have to add the MOAIColorGrid.cpp and MOAIFloatGrid.cpp (and .h) files to the various projects (xcodeproj, etcetera). I did this locally for xcode, but don't have the other tools to test against any of the others.

The documentation is NOT correctly updated, but I don't really know how to do that correctly.


makotok:
I think this fix and inflexible.
It is not a clean implementation.

I think those who have modified the MOAIGridDeck2D is good.
Or, create a new class that inherits from MOAIGridDeck2D.
That way, fix of MOAIGrid and MOAIProp is not required.

makotok:
I try to implement to try.

seebs:
I am aware that it's not a clean implementation. The problem is, I can't make any progress until I have a way to use the grid structure for something that needs per-item coloration. And while this implementation is cluttered right now, I think the basic idea of generalizing the grid to factor out the special cases like flags, and support other data types, is probably the right change.

I do see some point to maybe doing something like this in GridDeck2D instead of Prop, but even then, we'd still have a lot of reason to want the template-structured Grid, so we can use the same data structure, and mostly the same code, for multiple data types, not just ints which happen to have a few bits reserved to hold special flags.

It might make sense to keep the float/color/int grid implementation, but to do the color and scale stuff in GridDeck2D instead of Prop. Although really, it's not obvious to me that it's beneficial to have the grid logic in both Prop and GridDeck2D; perhaps it should be one or the other, rather than both.

superquadratic:
@seebs: Is this PR still valid, or is it superseeded by #789?

seebs:
I think 789 supersedes it.

Issue #779, opened 2013-09-13
Fix incorrect Untz enable macros 


CircuitCipher:
Fix the bug in android and possibly other hosts in which it uses the wrong preprocessor macro to enable Untz.

Seethe discussion here: http://getmoai.com/forums/android-untz-initialize-problem-t2053/

PR #783, opened 2013-09-21
Fix sound on iOS and Android host due to wrong macro

CircuitCipher:
Replace USE_UNTZ and USE_FMOD macros with MOAI_WITH_FMOD and
MOAI_WITH_UNTZ to fix sound on iOS and Android hosts.


CircuitCipher:
This addresses #779 which was brought to my attention by the forum posts here:
http://getmoai.com/forums/android-untz-initialize-problem-t2053/
http://getmoai.com/forums/moaiuntzsystem-on-android-t2071/

makotok:
This fix did not work well.
I want you to refer to this header file.

Definition of the macro would have been always.

https://github.com/moai/moai-dev/blob/master/src/config-default/moai_config.h

``` C
// moai_config.h
    #ifndef MOAI_WITH_FMOD_EX
        #define MOAI_WITH_FMOD_EX 0
    #endif
```

https://github.com/moai/moai-dev/blob/master/ant/libmoai/jni/src/moai.cpp

```
// moai.cpp
// NG
#ifdef MOAI_WITH_FMOD_EX
#include <moai-fmod-ex/host.h>
#endif

// OK
#if MOAI_WITH_FMOD_EX
#include <moai-fmod-ex/host.h>
#endif
```

For now, I will fix in development for the Branch.
After that, your root g merge the master.

makotok:
I was corrected.
I ask you for a check.

https://github.com/moai/moai-dev/pull/786

PR #786, opened 2013-09-23
Fix sound on iOS and Android host due to wrong macro

makotok:
I merged into the development branch from the master.
I fix because a collision occurred some.
Also, I fix together because there was a part that does not work.


CircuitCipher:
Tested on iOS and Android.  Sound works on both platforms.

PR #788, opened 2013-09-24
Fix for shaders and VBOs breaking on game resume

JDaance:
This is ex's fix from http://getmoai.com/forums/post7439.html#p7439

People are still running into this big issue. The fix is working well for me since about a year, I can only assume it's good.

### DIFF
--- a/src/moai-sim/MOAIShader.cpp
+++ b/src/moai-sim/MOAIShader.cpp
@@ -690,21 +690,15 @@ void MOAIShader::OnCreate () {
 		return;
 	}
 	
-	// get the uniform locations and clear out the names (no longer needed)
+	// get the uniform locations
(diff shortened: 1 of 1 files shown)


patrickmeehan:
We're wondering if this is still a bug. We use shaders in our own games on Android and haven't noticed a problem on resume. Wondering about this commit as it disables zglDeleteShader. Maybe it's just line 699 that's needed?

JDaance:
Well that sounds reasonable, I have no idea why the fix looks like it does. I assumed it was still a problem since it was reported on the forum.

 I'll be good and test it with and without the delete stuff and without the fix at all and see what happens.

JDaance:
I'm too lazy to test this right now since I'm a bit aways from a functioning build, if I run into it again I'll repost. Meanwhile I'll close this so as not to pester you with lingering pull requests :)

PR #789, opened 2013-09-30
Seebs/grid dev 3

seebs:
Okay, this is another try at making a usable color/scale kind of grid.

Step 1: Add an optional color argument to Draw() calls.
Step 2: Implement it for a couple of likely draw types.
Step 3: Move the logic for DrawGrid() from MOAIProp into MOAIGrid.
Step 4: Create a new MOAIGrid subclass which has a slightly more powerful draw routine, and also allows specifying additional traits.

The new design uses palettes rather than inline arbitrary colors largely for space reasons; it's pretty likely that, in many cases, a palette lookup of a smallish number of colors will be adequate, and it'll save a ton of space for a large grid.

I've found a couple of minor issues (a documentation typo), and provided hooks for setting the alpha channel separately per-item as well. This gets the functionality to a point where it can do most of what I was previously doing using Corona, and do it much more efficiently, too.


seebs:
I was really hoping to get this, or some variant of it, into the main tree some day; is there a specific problem that needs to be addressed, or was it just because this has been sitting around untouched for so long?

Basically, the ability to do things to individual items in a grid is really important for one of my major use cases. I'm willing to write more code, different code, better code, or whatever else, but I need some kind of feedback to act on to make whatever improvements are needed.

src/moai-sim/MOAIGridFancy.cpp
```diff
@@ -0,0 +1,559 @@
...
+			u32 idx = this->GetTile ( wrap.mX, wrap.mY );
+			u32 color = this->GetColor ( wrap.mX, wrap.mY );
+			float scale = this->GetScale ( wrap.mX, wrap.mY );
+			float alpha = this->GetAlpha ( wrap.mX, wrap.mY );
+			
+			MOAICellCoord coord ( x, y );
+			USVec2D loc = this->GetTilePoint ( coord, MOAIGridSpace::TILE_CENTER );
+
+			if (color) {
+				deck->Draw ( idx, remapper, loc.mX, loc.mY, 0.0f, tileWidth * scale, tileHeight * scale, 1.0f, this->GetPalette ( color ).ScaleAlpha ( alpha ) );
```
halfnelson, line 553: Instead of having a different deck->Draw for with color and without, couldn't you get away with using the GfxDevice:SetPenColor which would have the same effect (updates the final color which is applied already by the WriteFinalColor4b in the writeQuad functions. Would save a bunch of overloads and overrides?

seebs:
As I recall, the primary reason I didn't do that was that it would prevent you from _also_ using changing the color of the object as a whole. My intent was to have it so that you could have multiplicative color choices between "object as a whole" and "individual cells".

makotok:
I'm sorry, I closed by mistake.
I have created a merge branch. 
To merge once in 1-5.development branch if there is no problem.

https://github.com/moai/moai-dev/tree/future-grid-color

In addition, I was brought back code that is an obvious problem. 
Copy and paste code to reduce the maintainability. 
In addition, you should be using SetPenColor.

seebs:
I'd been thinking I should not use SetPenColor, because that would impose unwanted limitations and additional computation.
I suppose I could do this by using GetPenColor, multiplying colors in the loop, and using SetPenColor for them, but that ends up imposing a significantly greater computational load, because every item in the grid then requires re-multiplication by the ambient color, and also the generation and storage of both the 32-bit and color vector values.

makotok:
Thank you for your pull request. 
I was merged into 1-5.development branch. 
I think later additional fixes. 

In the original code, it seemed to only be used MOAITileDeck2D. 
By additional modifications, it should be used in all classes that inherit the MOAIDeck.

Performance will be required verification.

halfnelson:
seebs: instead of setPenColor, could we expose mFinalColor as a GET_SET instead of a GET and pull your final color calculation out of the writequad functions and into the grid draw loop? This would have almost the same performance profile and avoids the extra multiplications (you would have one or two more 32bit storages per grid cell, but the compiler should optimize that. If not, then no biggie.

Did you manage to get it so that you didn't have to put draw logic inside grid?

seebs:
I don't remember about the draw logic now, I haven't looked at it for a while. I am not sure whether just writing to mFinalColor would work, but I think it might. At that point, I could get mFinalColor at the top of the loop, do the multiplications, then restore it when done, and that would work.

seebs:
... Thinking about it more, though, I sort of dislike having the code know about and manipulate mFinalColor. Conceptually, I like the _option_ of a draw function that uses a color parameter, I just recognize that it may be a little intrusive. It seems like it might be useful to some other components occasionally, though.

halfnelson:
Fair enough. Just putting it out there. Great work BTW , nice clean code

naturally-intelligent:
Glad to see this getting attention, look forward to trying it out!

seebs:
I feel I should point out: The unmerged commits are not particularly relevant, the functionality seems to work just fine.

PR #792, opened 2013-10-04
add showCursor() and hideCursor() functions to MOAISim

atis--:
When developing something game-like on a PC system, it is usually necessary to disable the system cursor. I've added MOAISim.showCursor() and MOAISim.hideCursor() functions so that this would be possible. Just so you know, they are copy-pasted versions of EnterFullscreen and ExitFullscreen, This change also affects the GLUT host which sets the AKU\* callbacks.

Please check it out!


halfnelson:
Looks good to me. Great work

PR #794, opened 2013-10-04
Windows: A desirable switch to freeglut (multi-touch feature)

atis--:
This references an older pull request that I was asked by makoto to remake:
  https://github.com/moai/moai-dev/pull/728

As for the glutMultiButtonFunc() calls producing build errors on Unixes -- do you know why this is so? As I see it, freeglut is supposed to be cross-platform, so there shouldn't be any problems. I guess perhaps you have an older version of freeglut on your system that does not have these functions defined?

Anyway, here's the **text of the original pull request**:

Moai uses GLUT which, according to the freeglut web page http://freeglut.sourceforge.net/, has not been updated in over a decade. That, by itself, may not be reason enough to switch to freeglut. However, I believe there is a real and immediate benefit to doing so!

Not only is freeglut a drop-in replacement for GLUT, in its later releases it adds Windows multi touch support (WM_TOUCH). This is great news for those who happen to have a touch sensitive monitor under Windows (and possibly other desktop systems as well). One can then develop touch-enabled applications directly on the desktop (faster prototyping).

This pull request adds code to src/host-glut that employs freeglut's new multi-touch functionality. Switching all the project files properly from GLUT over to freeglut is beyond my competence, but code-wise nothing else needs to change: freeglut just works.

Regards,
Atis

### DIFF
--- a/src/host-glut/GlutHost.cpp
+++ b/src/host-glut/GlutHost.cpp
@@ -61,7 +61,7 @@
 
 #ifdef _WIN32
 
-	#include <glut.h>
+	#include <freeglut.h>
 	
 	#if MOAI_WITH_FOLDER_WATCHER
 		#include <FolderWatcher-win.h>
@@ -97,6 +97,7 @@ namespace GlutInputDeviceSensorID {
 		MOUSE_LEFT,
 		MOUSE_MIDDLE,
 		MOUSE_RIGHT,
+		TOUCH,
 		TOTAL,
 	};
 }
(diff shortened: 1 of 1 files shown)


atis--:
Maybe, same as in Windows, you have to include &lt;freeglut.h&gt; instead of &lt;glut.h&gt; to get the new functionality that is not present in the original GLUT library.

bullno1:
Recently, the host switched to SDL (2?) so I think multi-touch is covered.

atis--:
Judging from the source, SDL host is still in a more unfinished state than the GLUT host though.

superquadratic:
It looks like freeglut [does not support OS X](http://freeglut.sourceforge.net/progress.php). That is a dealbreaker. Also, as @bullno1 pointed out, the GLUT host will be replaced by an SDL host on the long run. So, unfortunately, I don't see a way to merge this change.

atis--:
Sounds cool, SDL is ultimately the better choice anyway.

Although for the sake of accuracy, I would like to point out that the linked page implies that freeGLUT does run on MacOS X through the use of XQuartz or similar.

PR #795, opened 2013-10-12
MOAIBox2DBody 'setGravityScale' and 'getGravityScale' functions added

stuartbnicholson:
Box2D has two useful functions: 'SetGravityScale' and 'GetGravityScale' on each b2Body instance. For details of their usage see the Box2d documentation.

MOAI does not expose either of these functions to Lua. This pull request is a minor change to MOAIBox2DBody that adds two new functions and exposes them to Lua.

1) getGravityScale returns the number representing the current gravity scaling of the b2Body. This is normally 1.0.
2) setGravityScale takes a single number and sets the b2Body's gravity scaling to that value. Any value is acceptable, but typically the -1.0 to 1.0 range is used.


makotok:
There is no fixed relationship is included.
If you do a pull request, do not fix in the master branch, please create a patch branch.

PR #796, opened 2013-10-28
Xcode 5 updates (quite possibly incorrect, I'm afraid)

seebs:
This got me able to build again after an unintended update to Xcode 5. This probably needs to be revisited by someone who actually understands how curl is getting configured and built, because I'm about 99% sure curl can be used successfully with 64-bit hosts.

But I figure putting the code out there to be looked at is better than not.

### DIFF
--- a/3rdparty/curl-7.19.7/lib/strequal.h
+++ b/3rdparty/curl-7.19.7/lib/strequal.h
@@ -31,6 +31,5 @@
 #ifndef HAVE_STRLCAT
 #define strlcat(x,y,z) Curl_strlcat(x,y,z)
 #endif
-size_t strlcat(char *dst, const char *src, size_t siz);
 
 #endif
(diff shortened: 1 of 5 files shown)


seebs:
I am not sure what is actually intended to be accomplished by this. In general, user apps should never be trying to provide their own declarations of library functions, and strlcat() is in reserved namespace. My guess is that the intent is that if the host system doesn't have strlcat, curl #defines it into something else and replaces it, and otherwise just provides a declaration which in principle matches the library's, except that OS X is doing fancy stuff that makes that declaration clash.

Upstream curl already fixed this:
http://sourceforge.net/p/curl/bugs/1192/

Their fix was just to drop the two uses of strlcat() in the code so they didn't need to solve this problem.

superquadratic:
Thanks for the pull request. Based on your work I created a new one, #800.

PR #798, opened 2013-11-06
Feature: CMake MOAI_ROOT variable

nanocell:
I added a MOAI_ROOT variable to the CMake build scripts, and changed all the source paths to be relative to this variable as opposed to having paths relative to the build scripts themselves. This will prevent the source paths from breaking if the build scripts get shuffled around (or moved in/out of subdirectories). This also allows the build directory to reside in an arbitrary location.


superquadratic:
Great, thanks! Merged into development branch in 42bde25e9b12a3219fbdef06d34a68c671d91004.

PR #801, opened 2013-11-16
Fix CMake build on OS X

superquadratic:
- Expat and Freetype did not compile
- There were linker errors regarding curl and crypto
- Linking did not work when generating an Xcode project


superquadratic:
Since I have the feeling that I'm the only maintainer interested in CMake on OS X, I'll merge this on Monday, unless there are any objections.

PR #805, opened 2013-11-23
Fix PVR texture data leak on iOS

superquadratic:
When a PVR texture was loaded (from file into memory) but never created (as an OpenGL texture), the texture data was never freed.


superquadratic:
Btw, once this fix is merged, I have some more changes ready that simplify the handling of compressed texture data and add support for ETC1-compressed textures on Android.

makotok:
Thank you!

PR #808, opened 2013-11-26
Fix Curl

superquadratic:
As reported [here](https://github.com/moai/moai-dev/commit/7ddf061485d54828320f1a160aa7478e5021a427#commitcomment-4701278), Curl does not compile (at least on Ubuntu) at the moment. That is because [this patch](http://curl.haxx.se/mail/lib-2013-02/att-0214/0001-strlcat-remove-function.patch) hasn't been applied correctly in 7ddf061485d54828320f1a160aa7478e5021a427.


superquadratic:
Merging since this was also reported in #811.

PR #811, opened 2013-11-29
Fix backport patch to remove strlcat from curl. 

cybergrind:
Accidentally leaved call of strlcat


superquadratic:
Thanks, but this is a duplicate of #808. :)

PR #812, opened 2013-12-05
Added text snapping to viewport scale as a setting on textbox

JDaance:
So this cleans up the change from #799 with a setting on textbox instead of removing the snapping altogether. It defaults to true so no changed behavior should be noticed!


makotok:
Thank you for fix.

PR #787, opened 2013-09-23
Feature dump Sep 23; revised Lua binding, SDL, misc

patrickmeehan:
Sorry for the huge pull request. One of the challenges of the git workflow is that sometimes smaller pull requests on their own branches can't get merged in before I've taken back into my own development branch and continued work. It's not feasible for me to have a bunch of little feature branches waiting around - there are often inter-dependencies. So it's going to be monolithic pull requests more often the not.

This one is going to break some stuff. Here's a quick rundown of the breaky parts:
- I'm tired of *glut and moving to SDL. Decided to build SDL as a static lib which is supported on OSX but not so much on Windows. To that end I commented out a big chunk of SDL_stdlib.c. I'm sure there's a better solution through compiler settings and header substitution; I just didn't have time.
- The new SDL host is _minimal_. I'll add more features as I need them. Hoping the community will fill in the rest.
- Have not tried running the revised test runner host. Not sure if this was ever added to CI or used for anything.
- I built Xcode OSX but didn't have time for the iOS sample project. Am hoping someone volunteers to fix it up.
- I changed the naming scheme for the modules to clarify the initialization/finalization process. It used to be that 'app' initialization was implicit, but this led to asymmetry of calls in the host (even though it handled all cases). This bugged me, so I made it more explicit. Each module now has an app initialize/finalize and a context intitialize/finalize. (Didn't have time to do the mobile modules - just desktop/windows so far).
- I've added some helper files for the 'official' Moai host modules. They are under the 'host-modules' folder. This is still a little half-baked, but the idea is to make some of the host tedium reusable and hide a lot of the ugly #ifdef's.

Some important fixes/changes:
- I did a lot of work on the Lua binding as it pertains to GC. Still not sure if we are out of the woods, but the intermittent crashes I was seeing in out game have gone away.
- MOAISim now takes control of the GC when it starts running. So business as usual until the first sim loop, then collectgarbage gets replaces with a warning and MOAISim starts running the GC incrementally. This let me get rid of a bunch of release/retains in the dep graph and action tree update code.
- The engine-to-Lua lifecycle was improved and now correctly provides a ref table to singleton classes so Lua objects created inside of constructors should now be exposed to Lua correctly.
- I added a way to set a custom GC callback on Moai objects for debugging purposes (dangerous - use with care). Also added a flag on MOAILuaRuntime to log GC events.
[...]

PR #818, opened 2014-01-12
Merge of patricks pr787 (lua_gc fixes etc) and cmake mods (work towards 1.5?)

halfnelson:
A shot at getting a stable 1.5
- This is a combination of https://github.com/moai/moai-dev/pull/787 plus all the fixes needed to get it running again and all hosts building again across all platforms. 
- It includes the html5 host that has been used by a few people now and runs the full set of hanappe samples
- It also includes updates for cmake build system to build ios, android, windows and html5 hosts.
- This has minor breaking changes to the lua side (the garbage collection call is now ForceGC) but custom host modules will need modification to talk to the new aku module interface and lua_gc references
- I have squashed the commits and split them into 2. 1 is all the new 3rdparty libs, the other is all the engine mods 
- glut hosts now use freeglut (except macosx)
- this can be resubmitted as a full merge if wanted (no squash) but thought this would be easier since there were so many commits and files updated.
  
  NOTE: I didn't touch the included project files. I focused on CMake. So a little help updating the project files if you want to maintain their current structure, or I can just use cmake to generate new project files for each platform (except the java one)


lordmortis:
One thing I'd seriously suggest looking at is including some of the 3rd party libs as submodules. You can tie a specific repo version to a specific submodule version (at least with Git submodules) and that means it should be less messy to upgrade from one submodule to another.  

I don't know when a good time to start looking at this is though...

Vavius:
Submodules are pretty messy when collaborating. If someone forgets to make "git submodule update" and then pushes to the repo we will have this submodule rolled back. 
It's too easy miss a single line where submodule hash changes.

lordmortis:
That's true, but there's a lot of places where similar things could happen. Perhaps an initialisation script that knows what the current versions of submodules are and the CI system uses that to update properly before building? (and CMake can use this to set submodules up?)

halfnelson:
It would also mean that you would need 2 pull requests for updates to 3rdparty libs, one for the lib and one for the moai code. Some repos will need to be forked and kept under moai since zipline have made minor mods to their source.

halfnelson:
I like the idea of a cleaner way of handling 3rdparty libs, but I also think that there is something nice about knowing that a single checkout or download of zip will have all of the required code to build and that the commits are kept together so it never gets out of sync. I think Patrick prefers it without submodules

Vavius:
Flower sample crashes. Somebody should reapply recent changes to xcode project (mainly luajit), I've never used luajit.

Vavius:
Halfnelson, I think we should change flower, not moai. I've looked through changes and it's just a new API. 
Flower fix is one-liner. 

```
-- before it was like that:
obj:setInterface(self)

-- and now we should use it like this:
obj:setInterface({__index = self})
```

Vavius:
Doesn't work, must be I mixed up my moai executables.

halfnelson:
I have found a way to fix the problem, I think it is actually a bug in moai, not a misuse. The search for missing function is meant to go `userdata->membertable-interfacetable`
but actually goes
`userdata->membertable->interfacetable.__index`

If I patch up the metatable/__index chain so that interfacetable is included in the search then flower works again.
downside is that with current moai this needs to be done after each call to setInterface or setMembers

see
https://gist.github.com/halfnelson/8412713

Vavius:
There is an example of new behaviour in samples/test/lua-binding. 
setInterface now just sets the metatable for member table and nothing more. So now we should explicitly set __index of interface. 

Something like this should work: 

```
newInterface = {}
setmetatable(newInterface, newInterface)
newInterface.__index = MOAIProp2D.getInterfaceTable()

prop = MOAIProp2D.new ()
prop:setInterface({__index = newInterface})
```

The strange thing is, after applying this change to flower yesterday at my work mac everything worked fine, but here at home flower segfaults after the change. While sample above is fine.

halfnelson:
Yep good find vavius. That is indeed what is happening. I am not sure if this was by design? or just an oversight. It seems a strange use of setinterface, to set a metatable!?

i have the following that gets flower working:

``` lua
        local interface = { }
        setmetatable(interface,interface)
        interface.__index = self
        obj:setInterface(interface)
```

and the flower samples now run fine. (I also added MOAISim.forceGarbageCollection = MOAISim.forceGC for good measure)

halfnelson:
turns out we can get flower running by prepending:

``` lua
     local oldfunc =  MOAISim.setInterface
     MOAISim.setInterface = function(self, newInterface)
          local interface = { }
          setmetatable(interface,interface)
          interface.__index = newInterface
          oldfunc(interface)
        end
```

Vavius:
I think I've merged xcode projects... 
Now they compile and I can see anim sample on my iPhone. 

I've took projects from current master state and added changes above. OSX host is now SDL-based. Xcode produced a moai binary where I can't resize window, so it's different from old glut. 

here is the commit https://github.com/Vavius/moai-dev/commit/14a6a340f5ab90906c4cdbe76607c56e905dee58

Managing xcode projects is a nightmare. We should really deprecate huge libmoai project and keep only hand-crafted host projects for iOS and OSX, which will include cmake-built static moai libs.

halfnelson:
great work, i will merge it into this pull. I agree regarding libmoai project, but i would keep the moai-iphone library project (has all the extensions) as handrolled too. The cmake code generates buildable xcode projects that you can open and run from in xcode.

halfnelson:
Update pull request to not be a squash of patricks changes but just a normal merge. This will help in keeping track with patrickmeehan/development in the future as it remembers each commit already processed.

makotok:
Merges to be able to build in osx, I have created a branch. 
Hanappe may also work. 

To be merged into development branch if there is no problem.

PR #823, opened 2014-01-26
Android print() fix

quocble:
This is the most elegant solution I found.


iamricard:
Works for me too. Thanks for the quick hack!

PR #828, opened 2014-02-24
Hiq camera feature support

operat0r:


operat0r:
added camera support for the android host, we can now start the camera app from within moai, take a picture and use it inside moai.

ios support wil follow soon

operat0r:
also changed bash build scripts - now using all available cores
enabled logging on android host

makotok:
Thank you for the new feature.

PR #833, opened 2014-02-25
Add Box2D closest raycast support

naturally-intelligent:
Adding closest raycast support, as taken from the following thread/post:
- http://getmoai.com/forums/raycast-for-box2d-hack-t562/
- http://www.fivevolthigh.com/2013/10/moai-exposing-box2d-world-ray-casting-to-moai/

This feature is working well and complete in its own right but Box2D does have additional raycasting functionality that should be exposed too, via a callback to test if raycast should continue on through a fixture or not.  However even without that callback this closest raycast is still extremely useful.

### DIFF
--- a/3rdparty/untz/src/native/openal/UntzSystem.cpp
+++ b/3rdparty/untz/src/native/openal/UntzSystem.cpp
@@ -55,7 +55,7 @@ static inline SInt16 limit_float_conv_SInt16(float inValue)
 }
 
 void *audio_loop (void *ptr) {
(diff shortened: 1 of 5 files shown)


naturally-intelligent:
Umm not sure why it has now added my Untz changes, I did those after I sent this pull request.  Also sorry this is targeted at master not development.  Sigh.  Will try and fix.

naturally-intelligent:
Well I will leave it Open in case you decide to use it, it's still good, but I can't figure out how to undo with Git etc.

PR #834, opened 2014-02-25
Hiq numeric keyboard support

operat0r:
added support for number-, datetime, phone-keyboard on android

fixed critical bug on MOAIKeyboardAndroid::_hideKeyboard (segfault because of wrong java method lookup - strange that this worked until now)
fixed return types for functions called from java


makotok:
Thank you for correction. 
Future, it will be added to the same interface MOAIKeyboardIOS.

PR #837, opened 2014-02-27
Pull cmake improvements

halfnelson:
Small fix to the doc generator's version parser
Improve visual studio support with and speed with parallel builds
Improve generated vs projects with solutions folders
fix curl build (needed zlib)
Clone sdl2 so we can edit the cmakelists.txt and added to solution folder


makotok:
Thank you for fix.

makotok:
This fix will I build error in the Windows environment?

http://getmoai.com/forums/error-building-vs-solution-via-cmake-t2531/

DanielSWolf:
@makotok I'm not sure what you're asking. As I wrote on that forum thread, it seems to me that this pull request (#837) somehow introduced the problem.

DanielSWolf:
Wrong alarm! The problem was actually caused by my virus scanner! Sorry about that!

halfnelson:
no problems. Does it build faster for you and are you happy with folder layout?

DanielSWolf:
Yes on both accounts! :-)

PR #851, opened 2014-03-03
SDLHost: Fixing Segfault caused by wrong handling in huge SDL keycodes

theypsilon:
Keycodes of keys without character representation are bigger than 0x40000000 (see http://wiki.libsdl.org/SDLKeycodeLookup). And keycodes in SDL2 are more than 258. 

This fix translates huge keycodes into minor ones, and extends the key storage array from 258 to 512.

### DIFF
--- a/src/host-sdl/SDLHost.cpp
+++ b/src/host-sdl/SDLHost.cpp
@@ -142,10 +142,13 @@ void MainLoop () {
 					break;
 				
 				case SDL_KEYDOWN:
-				case SDL_KEYUP:
-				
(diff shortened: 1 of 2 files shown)


makotok:
No problem with GLUT version?

makotok:
Merge the time being. I will check back later.

theypsilon:
Can't compile GLUT now on my Ubuntu 13.10 (linking errors). Long time ago, it was running ok, but most of the keys without character representation (i.e.: the direction arrows) didn't work at all, they were not being registered in the eventStream.

PR #854, opened 2014-03-05
Support Android url scheme and event APP_OPENED_FROM_URL

eiriksm:
This adds the event APP_OPENED_FROM_URL to the android host as well. Very useful for deep-linking. We are using this in our host, but it presents one problem:

The code included here in the pull request will only trigger the event if the app is already running. As this very often is not the case, we hacked around it by triggering the event a little bit delayed in the onResume function. If anyone with android skills and knowledge have a good solution to this, that would make the pull request more complete.

### DIFF
--- a/ant/host-source/source/project/AndroidManifest.xml
+++ b/ant/host-source/source/project/AndroidManifest.xml
@@ -27,6 +27,13 @@
 				<action android:name="android.intent.action.MAIN" />
 				<category android:name="android.intent.category.LAUNCHER" />
 			</intent-filter>
+			<intent-filter>
+				<action android:name="android.intent.action.VIEW" />
+				<category android:name="android.intent.category.DEFAULT" />
+				<category android:name="android.intent.category.BROWSABLE" />
+				<!-- Add your URL-scheme here -->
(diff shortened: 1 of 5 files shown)


eiriksm:
Wow, thanks! That was quick!

However, we should still solve the issue about getting the event to trigger if the app is not already running. Do you want me to open a separate issue for it?

makotok:
If there is another problem, I want you to pull request. 
At least, I seem to be no impact on existing code. 

Correct any problems.

eiriksm:
As I mentioned, this adds the event for APP_OPENED_FROM_URL on android. However, it will not trigger if the app is not already running.

For example, if my app registers the url scheme eiriksm:// and i open the app, and after that i open the url eiriksm://awesome?test=true, the app will open, and the event will trigger. But only if the app is already running. If it is not running, the app will open if I open the url eiriksm://awesome?test=true but the event will not trigger.

I will try to find a better solution to this than the one we are already using in our game, and post a new pull request. But if anyone is reading this and have a clear idea on what to do, please feel free to give me any hints :)

makotok:
Oops. Compile error occurs.

Issue #857, opened 2014-03-10
Fall to the keyboard at the end of Android.


makotok:
The error in the keyboard at the end of Android.
Any modifications are affected.

halfnelson:
give that pull request a go. It should fix the problem

PR #859, opened 2014-03-15
Pull fix android back button

halfnelson:
this fixes a crash at the end of the android app where it was trying to access an already freed callback.
This should fix https://github.com/moai/moai-dev/issues/857


makotok:
Thank you Fix!

Mattenklopper:
This might need another change, see: http://getmoai.com/forums/android-1-5-back-button-crash-on-4-4-t2561/#p13630

makotok:
@Mattenklopper
Thank you for information. 
It should not be commented out this code. 

This is to determine whether to cancel the back button in Lua side. 
Therefore, the listener side MUST return a True / False.

Mattenklopper:
I am not very experienced with the android <-> moai development.
Something has to be changed though, as right now when you create a custom back_button function it gets called multiple times and on both action_down and action_up.

PR #863, opened 2014-03-17
Added methods for dumping the Lua stack to MOAILuaState

DanielSWolf:
These methods are handy when debugging Moai's low-level Lua calls.

### DIFF
--- a/src/moai-core/MOAILuaState.cpp
+++ b/src/moai-core/MOAILuaState.cpp
@@ -421,6 +421,57 @@ void* MOAILuaState::GetPtrUserData ( int idx ) {
 }
 
 //----------------------------------------------------------------//
+STLString MOAILuaState::GetStackDump () {
+	STLString out;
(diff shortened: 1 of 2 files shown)


src/moai-core/MOAILuaState.cpp
```diff
@@ -805,6 +856,18 @@ bool MOAILuaState::PrintErrors ( FILE* file, int status ) {
 }
 
 //----------------------------------------------------------------//
+void MOAILuaState::PrintStackDump () {
+	STLString stackDump = this->GetStackDump ();
```
jeremyong, line 861: the `this->` is superfluous here and in the method below no?

DanielSWolf, line 861: Yes, it's not required. I saw the method below (PrintStackTrace) using the explicit syntax and assumed it might be a Moai coding style thing.

PR #864, opened 2014-03-17
Fixed bad upvalue for MOAILuaClass::_extendSingleton()

DanielSWolf:
`MOAILuaClass::InitLuaSingletonClass()` tries to create a C closure from the `_extendSingleton()` method with the userdata and class table as upvalues. Even after @halfnelson's fix, however, it instead used two copies of the class table as upvalues. As a result, `_extendSingleton()` always received a NULL-pointer as first upvalue (because a table cannot be converted to userdata).


makotok:
Anyway, I merged. 
Fixed in additional if there is a problem.

PR #867, opened 2014-03-19
Pull html5 improvements

halfnelson:
I have done some refactoring of the moaijs host. It uses a new file packager that outputs json, it caches data file in indexedDB. It can be simply embedded as an iframe (helps with blog posts and documentation).
I have also added a build-html-win.bat for building the emscripten host on windows.


makotok:
Thank you!

PR #869, opened 2014-03-25
remove romloader,js

ezraanderson:


ezraanderson:
halfnelson. I think this file needs to be removed

PR #870, opened 2014-03-25
fix linux untz build problem

halfnelson:
reduce cores for linux build to prevent errors on older boxes
update linux glut build script to include luaext and http client.


makotok:
Thank you for fix. 
In Linux, even GLUTHost, SDL was needed?

halfnelson:
yes. the linux glut host uses sdl for untz, the **SDL** define is only for
untz (as far as a code search can tell)

PR #877, opened 2014-03-31
Fixed disabling of ios host modules

JDaance:
Modules didnt get properly disabled by script

Facebook didn't get disabled in code

Mobile App Tracker didn't have disable option


makotok:
The following code assumes the first argument.

```
LUASRC=$(ruby -e 'puts File.expand_path(ARGV.first)' "$1")
```

makotok:
To merge. 
Fixed in added if it does not work.

Issue #881, opened 2014-04-07
MOAIHttpTask not exposed in LUA with Windows host


romainmoreau:
I tried building with MinGW with bin\build-win.bat with -DMOAI_CURL=TRUE.
I also tried generating projects with bin\create-projects-win.bat with -DMOAI_CURL=TRUE and building with VS 2013.
In both cases, I get this error on the LUA side when I try to use MOAIHttpTask : attempt to index global 'MOAIHttpTask' (a nil value))
Also, in both cases, I don't see in the build log a line about compiling MOAIHttpTaskCurl.cpp.

halfnelson:
Nice find. The build-win.bat is missing a `-DMOAI_HTTP_CLIENT=TRUE` flag. Add this as well as ensuring that `-DMOAI_CURL=TRUE` and it should work. If this fixes it could you submit a pull with the changes. Thanks

romainmoreau:
Thanks for the quick update, your fix works, I just submitted the pull request.

makotok:
Merged.

PR #882, opened 2014-04-07
Fix wrong path and missing HTTP Client in Windows scripts

romainmoreau:
Please see the issue : https://github.com/moai/moai-dev/issues/881
This fix also activates OpenSSL for HTTPS and fix a path issue.


bin/build-win.bat
```diff
@@ -1,6 +1,6 @@
-cd cmake
+cd ..\cmake
```
halfnelson, line 1: This wasn't wrong path. Docs say to run this script from root of sdk like: bin\build-win.bat which needs cd cmake

romainmoreau, line 1: You are absolutely right, I reverted the change and submitted another pull request.

PR #875, opened 2014-03-29
Removed Box2D 2.2.1, and replaced it with Box2D 2.3.0. Changes to b2Worl...

fivevolthigh:
...d and MOAIBox2DBody to support Box2D 2.3.0.

These are changes to move 1.5-development to Box2D 2.3.0, prior to attempting to integrate LiquidFun into MOAI, as LiquidFun is based on Box2D 2.3.0.

Issue #871, opened 2014-03-26
Warning of an inline function.


makotok:
When building in XCode5, a large number of alerts occur.
It is already fixed in the repository of Patrick.

```
/Users/mkinoshita/moai-dev/src/zl-vfs/zl_replace_new_delete.h:59:2: warning: 
      replacement function 'operator delete[]' cannot be declared 'inline'
      [-Winline-new-delete]
        inline void operator delete []( void* ptr ) throw() {
```

makotok:
zl_replace_new_delete.h

This header is necessary? 
Patrick has been commented out already.

I have corrected the warning, but it is an error that it is OSX(XCode 5).

```
Linking CXX executable moai
duplicate symbol __Znwm in:
    CMakeFiles/moai.dir/Users/mkinoshita/moai-dev/src/host-sdl/SDLHost.cpp.o
    CMakeFiles/moai.dir/Users/mkinoshita/moai-dev/src/host-sdl/SDLHostMain.cpp.o
duplicate symbol __ZnwmRKSt9nothrow_t in:
    CMakeFiles/moai.dir/Users/mkinoshita/moai-dev/src/host-sdl/SDLHost.cpp.o
    CMakeFiles/moai.dir/Users/mkinoshita/moai-dev/src/host-sdl/SDLHostMain.cpp.o
duplicate symbol __Znam in:
    CMakeFiles/moai.dir/Users/mkinoshita/moai-dev/src/host-sdl/SDLHost.cpp.o
    CMakeFiles/moai.dir/Users/mkinoshita/moai-dev/src/host-sdl/SDLHostMain.cpp.o

...
```

PR #885, opened 2014-04-09
Seebs/warning fixes

seebs:
This is a handful of fixes to remove warnings being generated by clang. I think at least a couple of these were serious bugs that could break code badly on some platforms (such as anything where size_t isn't the same size as unsigned int). Mostly just trying to reduce the spam so that warnings are actually significant.


makotok:
Thank you for fix.
I hesitate to merge.

1)
You have to modify the source of a third-party. 
Is no longer the original. 

The future, problems will occur when it is upgraded.

Can not suppress a warning in the compilation options?

makotok:
I would like to seek the opinion of other people.

seebs:
These are really good warnings to have, and at least a couple of them are for code which will actually fail on some devices, probably. (But possibly not anything most people target.) So I wouldn't want to _generally_ disable these warnings. I do recognize there's difficulties with maintaining modifications to the third-party code, and in at least some cases I think it makes sense to ship those upstream. Although with Untz, at least, I think that's also owned by the same people. I'd be willing to try to contribute some of these patches back to upstreams.

The inline qualifier change, and Untz changes, are local code only I believe, even though Untz is categorized as third-party. The lcrypto patch for the variable type is an actual real code bug that could be dangerous on some systems.

makotok:
I also think we should definitely fix some.
I want to merge if there is no objection in particular.

Vavius:
We shouldn't modify third-party stuff that way. 
We will get those warnings back when upgrade libs.

seebs:
So what? In the mean time we'll get fewer warnings, and since at least some of these warnings represent _actual bugs_, I think we should definitely fix the local copies. I am all for submitting fixes also to upstream, although in many cases I suspect they'll get fixed anyway. We aren't the only people who ever build with clang...

bullno1:
I agree with Vavius. We shouldn't modify third-party stuff. Warnings are _potential_ bugs. However, third party libraries are thoroughly tested by other projects too. They should be relatively bug-free. Modifying them may actually _introduce_ new bugs.

We should only modify our own code.

For third party, we can either
- Upgrade them for less warnings. zlib is a likely candidate. The latest stable is already 1.3.5.
- Suppress warnings through compiler option

Vavius:
Third option: 
- Fix warning, publish pull request to third-party supplier, upgrade to new version

It is common practice to suppress third-party warnings. 

P.S. We can treat UNTZ as our code, not third-party

bullno1:
Then we need to make sure we are using their latest stable, which might not be the case.

Btw, I'm worried about some changes like:
https://github.com/moai/moai-dev/pull/885/files#diff-b9b8dcffbe29d5a31d7a95e2c5785369R22: I recall that without inline, MSVC will choke. Do we have CI for Windows or just Linux?

https://github.com/moai/moai-dev/pull/885/files#diff-316c76f75812d4b907a5f9d6ad5affedR24: change the layout of MOAICoroutine. It may be incompatible with future updates.

https://github.com/moai/moai-dev/pull/885/files#diff-3688334e94bb0a870cbc7d768b1532edR34: This assignment is not needed since the member is already initialized. I think it's a merge/refactor artifact.

https://github.com/moai/moai-dev/pull/885/files#diff-4b239dfc7b6d12d629c99584000201b8R60: Not sure what the warning for this is. Getting values from method and from member are two totally different things. It may break future code. Other changes in 3rdparty/untz/src/BufferedAudioSource.cpp has the same problem.

I'm fine with other changes to Untz and Moai

My point is: suppress through build option should be preferred, not modification.
Lots of stuff you commented out are just dead code which will be optimized away anyway. I don't think the risk of making a mistake somewhere in modifying third party code is worth it.

seebs:
I think it is a very bad idea for us to assume that we are completely incompetent compared to the people producing third-party libraries, but without that assumption, I don't see a reason to assume that we cannot possibly be right to modify their code. At least some of these issues are real bugs. Yes, it's a common practice to suppress warnings on third-party code, but we should at least evaluate them carefully first. But I don't see any reason to assume that, in general, this code is any more bug-free than Moai is.

For the inline/no-inline thing: That's a C++ language spec thing, inline _isn't allowed there_. If MSVC chokes without it, then we'll have to do a conditional, but usually they don't specifically require that the code be incorrect, do they?

MOAICoroutine: Changing the layout shouldn't matter, because we aren't relying on binary compatibility with memory dumps. And "future updates"? This isn't third-party code, we control the updates and can do whatever we want.

A lot of the BufferedAudioSource things are warnings because the code was requesting values and then not using them. This isn't a distinction between getting values from method and from member, it's a distinction between calling a method once and using the returned value several times, or calling the method several times. If you think it's better to call the method several times, we can just drop the initial store to a then-unused variable, but all of these were "variable assigned and then never looked at" warnings, which look like they came from someone starting to cache a result and then not updating other code to use that.

makotok:
There are things to be aware of some.
- Plan to upgrade to LuaCrypto 0.3.2.(#891)
- Box2D has been updated with 1-5.development branch.(#875)

Therefore, conflict occurs several fixes.

seebs:
The possibility of merge conflicts is why these are separated out per-package.
A thing occurs to me: In the build system I normally use, when we have third-party code, we separate that code out from the patches we apply to it before building, for cases where we need to fix a bug or warning and don't have time to wait for upstream to merge it and possibly release a new stable version.
It might make sense to change things to have a patch phase so that we have the original upstream sources, then during a build we copy them, and apply patches. Only I only know how to do this for Unix-like build machines, not sure what the options would be for Windows-like build machines.

makotok:
The point that I was worrisome.

1) Drop inline qualifier
- src/zl-vfs/zl_replace_new_delete.h

=> Better to fix would be nice.(#871)

2)  lcrypto: use unsigned int for functions taking unsigned int *
- 3rdparty/luacrypto-0.2.0/src/lcrypto.c

=> Conflict occurs.

3) xmplparse: Handle warning from clang
- 3rdparty/expat-2.0.1/lib/xmlparse.c

=> There is a possibility that the behavior changes. We should keep the compatibility. Or do not fix.

4) MOAIUntzSampleBuffer.cpp: Drop unused variable
- src/moai-untz/MOAIUntzSampleBuffer.cpp

=> Should be corrected.

5) tlsf.c: Don't do a dummy assignment.
- 3rdparty/tlsf-2.0/tlsf.c

=> Need?

6) Suppress warnings
- 3rdparty/luasql-2.2.0/src/ls_sqlite3.c

=> Can not be determined. Should not modify the better?

7) jansson: Try to reduce warnings.
- 3rdparty/jansson-2.1/src/hashtable.c
- 3rdparty/jansson-2.1/src/jansson_config.h

=>Can not be determined.

8) Fix various warnings: 
- 3rdparty/untz/src/BufferedAudioSource.cpp
- 3rdparty/untz/src/MemoryAudioSource.cpp
- 3rdparty/untz/src/OggAudioSource.cpp

=> Bug? Better to fix is good.

9) sfmt: only declare sse2_param_mask if using it
- 3rdparty/sfmt-1.4/SFMT.c

=> Can not be determined.

10) Warning/error cleanup
- src/moai-sim/MOAIBlendMode.cpp
- src/moai-sim/MOAIFrameBuffer.cpp
- src/moai-sim/MOAISim.cpp

=> Better to fix good.
- src/moai-sim/MOAICoroutine.h

=> Should really be modified?

11)     Remove unused variables from Box2D 
- 3rdparty/box2d-2.2.1/Box2D/Collision/b2TimeOfImpact.cpp
- 3rdparty/box2d-2.2.1/Box2D/Dynamics/Joints/b2GearJoint.cpp
- 3rdparty/box2d-2.2.1/Box2D/Dynamics/Joints/b2PrismaticJoint.cpp
- 3rdparty/box2d-2.2.1/Box2D/Dynamics/Joints/b2RevoluteJoint.cpp
- 3rdparty/box2d-2.2.1/Box2D/Dynamics/Joints/b2WeldJoint.cpp

=> I think should not fix is good. It is up to version Box2D 2.3.0 Whatever the case may be.

12) luacurl: Clean up a comment
- 3rdparty/luacurl-1.2.1/luacurl.c

=> Failure of the comment. However, it does not affect much.

makotok:
Multiple fixes are included. 
Can you split a pull request?

seebs:
I don't know for sure how, but I am pretty sure github lets you partially merge changes.
For things where there's significant upgrades to the package upstream, should we look at doing package uprevs first?

bullno1:
It's not about being less competent than them. It's about whose code is tested more out there.
It's not about we cannot possibly be right, it's about we can possibly be wrong.
I prefer known and documented bugs to unknown ones.
When I was talking about "future updates", I meant what comes from zipline's branch

Btw, I haven't seen any real bugs in the code.

BufferedAudioSource: Sorry, I wasn't looking at the code above. That's a totally valid fix. My bad.

I vote for update third-party first. Then we revisit these changes. Changes to Moai are valid anyway.

bullno1:
@makotok 
1) I need to check MSVC. I might be wrong.
2) Scary stuff right there. Update lcrypto first. For 32-bit, we should be fine but for 64-bit, all hells will break loose. We need actual tests for this. @seebs is this what you meant by "actual bugs?
3) Either find a better fix or don't fix. Just suppress.
4) Good to merge
5) Good to merge
6) char signedness. The compiler was already doing something implementation-defined implicitly. seebs just made it explicit. Should be fine.
7) hashtable.c: Probably unused const. If there's no link error, it's fine.
    jansson_config.h: Should be something to do with Objective-C++ compiler not defining __cplusplus. I'd prefer:

```
#if defined(__cplusplus) || defined(__APPLE__)
```

It makes the flow more obvious

8) Not real bugs. But valid warning fix.
9) No idea. Try upgrading first.
10) Good to merge. Initialization order should be in order of declaration.
The MOAICoroutine thing: I'm partial. We can remove it and add later in the future if it's really needed.
11) Upgrade first, fix later
12) Don't see the harm. Just merge.

seebs:
It's not that the Objective-C++ compiler doesn't define __cplusplus, it's that this is biting us when compiling as C, not C++. The problem is that we've got a statically-generated jansson_config.h, which has to work both with Apple's compiler (which does support inline, and complains about the unused functions if they're not inline) and other compilers (which may not support inline for C).

And the (size_t *)/(unsigned int *) thing is absolutely, positively, a real bug. That code is just plain wrong, and wrong in a way which can and will blow up on real hardware.

makotok:
Thank you for your feedback. 
I want to merge individually.

I'm currently working in the death march project. 
If it is not a simple task, time-consuming.

(I wish I have volunteers to someone else)

EDIT:
May I merge all? 
To resolve the conflict with the other needs to merge.

JDaance:
Please don't change any third party code if there isn't a _really_ strong reason to do so:) It might introduce bugs ( I know it's unlikely but just the fact that it could is enough), and it will be very messy to figure out which version of stuff we are running, and if any of our old changes have to be merged or carried over to a new version. By all means change untz and moais own stuff though!

If we really have to change a third party lib I think we should rename it to like box2d-2.1.1-moaicustom with some kind of readme notice to know how to pay attention when upgrading or debugging issues.

seebs:
Of course it _might_ introduce bugs.
The question is whether it's more likely to introduce bugs or fix them. Also, whether the effects of the existing state are serious. Which they are. If you have >200 warnings from a project, you _don't know_ whether you are getting warnings for serious bugs. Some of those warnings are real live bugs that will result in catastrophic failures on some targets... Some might not be. We don't know until we evaluate them.
I would consider "you can't build the simulator or an app without warnings" to be a _really_ strong reason to fix things. Compiler warnings are often for things that will crash at runtime or produce severe errors including possible data loss.
I recognize the possible concern with figuring out which versions we have, but this is a well-studied problem we can address. My inclination would be to suggest a build process which takes upstream sources and applies patches to them, so we can clearly see where our patches are.

patrickmeehan:
Since there's still debate (and since we can't auto-merge) I am closing this one. Though it pains me!

Current policy is to turn warnings off for third party libs.

I will re-fix untz and moai warnings (or bring these fixes over by hand) when I have a moment.

The libcrypto warning/error is definitely alarming, so I'll double check that one as well.

PR #896, opened 2014-04-13
Call the lua callback from the java side, add default (required) variables for google play services.

eiriksm:
I am not sure what the norm is on calling these AKU functions, but at least this is tested and works.

Also added some variables to androidmanifest and strings.xml. These are required and would crash the app if not set.


eiriksm:
Yeah, I thought of that actually. Don't know why I did not add it.... Will do

eiriksm:
Added now, still attached to same pull request. Thanks for pointing it out!

makotok:
Thank you for the improvement.

PR #897, opened 2014-04-15
Remove community fork from readme

JDaance:
This IS the community fork, the old one is just confusing :)


DanielSWolf:
Good find! It might also be a good idea to delete the community fork altogether - there hasn't been any activity over there in eight months. @franciscotufro: What do you think?

JDaance:
Yeah I second that :) Please remove it

PR #902, opened 2014-04-18
A few small fixes

criptych:


halfnelson:
which version of mingw are you using. The whole system builds fine under mingw64. I would be reluctant to change third party code

criptych:
OK, yes, I'll have to double-check my version, though I think it's fairly recent.  `sdlchecks.cmake` is definitely broken, though -- or at least, it was when I made this PR, but I see it's been fixed already. :)

halfnelson:
yep. The ssl fixes should just be a warning in the latest mingw64

makotok:
Not be merged. 
The need to re-fix?

criptych:
I just freshly downloaded and installed it on a new computer, and had the same issue.  However, I think the problem may be that I'm using plain MinGW, not MinGW64.

halfnelson:
Yep. Mingw64 is recommended. It is what the scripts were developed for and has a bunch of bug fixes

criptych:
Using MinGW64 solves the OpenSSL issues, but I still get lots of errors related to `fileno` vs. `_fileno`.

criptych:
Never mind, seems like it was still using the configured MinGW include paths. Starting with a clean build directory, it builds fine.

PR #917, opened 2014-05-14
Fix MOAI zip file reading with 64bit systems, as well as small (< 256b) ...

ludamad:
...zip files.


ludamad:
Ideally instead of s/long/int/ it would be some uint32 type but I wasn't sure the preferred way for this to be done. As well, the code was previously segfaulting with small files. (This took quite some debugging headache.)

halfnelson:
https://github.com/moai/moai-dev/blob/master/src/zl-common/zl_types.h has
the typedefs for int sizes used across the zl base classes. Maybe it just
needs to pull in that file.

ludamad:
Thanks. I adapted and went ahead and clarified the 'unsigned short' usage too. Tested and works on my end.

ludamad:
Excellent, thanks for merging.

Issue #919, opened 2014-05-21
ios-host, lua-headers-wrappers, are broken 


ezraanderson:
https://github.com/moai/moai-dev/blob/7b0de45d8904ede794d7665cf44746e83895eca7/src/host-ios/Classes/MoaiView.mm

execution order is incorrect, prevent moai.lua (headers) values to be correct

 AKUModulesRunLuaAPIWrapper (); --> should be bellow --> AKUIphoneInit ( application );

halfnelson:
this has been merged. Thanks!

PR #920, opened 2014-05-21
Run the lua api init AFTER iphone init, fixes moaidialog etc not being i...

halfnelson:
...nitialized


halfnelson:
Should fix #919

PR #931, opened 2014-05-29
Code clean-up

DanielSWolf:
Removed some dead code in moai-core


makotok:
Thank you for correction. 
I hesitate to merge. 

Since the API has already been published, it may be someone in trouble. 
In addition, I look useful some.

DanielSWolf:
Hi Makoto,

Thanks for the feedback. Let's look at the individual commits.
- `MOAICast` and `RTTI`: These aren't used anywhere in Moai, but I agree that somebody else might be using them. So it's alright to leave them in.
- `MOAIScopedLuaState::PinTop()`: This method isn't used anywhere in MOAI, and I believe it is rather harmful. After all, the whole point of `MOAIScopedLuaState` is to restore the initial Lua state
  at destruction, not to change it.
- Retiring/restoring global classes: To me, this looks like an attempt to introduce Moai-wide state handling that was never finished. I don't believe there is any harm in removing this code.
- The remaining three commits remove leftovers that are without any use.

So in short, I'd recommend not merging the first two commits, but merging the rest.

patrickmeehan:
Do me a favor and leave those for now.

Removing the MOAICamera2D method bodies and declarations w/ no implementations is fine.

DanielSWolf:
Sounds good. @makotok: Can you merge my commits 2dd190e, cd34800, and 8f4d2ba ?

DanielSWolf:
I'm closing this pull request. I'll create a new one containing just the approved commits.

PR #942, opened 2014-06-16
Safeguard for systems not supporting STREAMS

monsieurh:
Recent Linux systems don't support STREAMS, but have the flag #ifdef HAVE_STROPTS_H leading to a compilation error when trying to reach <stropts.h>. The following safeguard tests explicitly the presence and support of STREAMS and uses <sys/ioctl.h> as a replacement.


halfnelson:
This has been fixed in master by changing the config header

PR #944, opened 2014-06-24
Latest updates on community integration

patrickmeehan:
This is the latest. I'm going to be working from my copy of this branch moving forward.

Android is going to be extremely broken for a bit.


bullno1:
@patrickmeehan About plumzi's untz changes: Is this going to include all the codecs or just selected one? If I'm not wrong, BlueTune is GPL, is that going to affect Moai or games using it?

patrickmeehan:
@bullno1 Not even aware of that stuff. Do you want to take point on the latest Untz? Definitely don't want to ship any GPL...

PR #955, opened 2014-07-20
work around incompatibility with lua_jit and the coroutine error handlin...

halfnelson:
...g functions

a runtime error in a coroutine sets the threads lua_state status field to 2. In luajit in debug mode there is an assert before each pcall that checks the current thread status. This causes our error handler to assert and crash project instead of printing stacktrace. I have written a patch that lets it just dump the stacktrace (the default method of handling errors) when running under luajit.

### DIFF
--- a/src/moai-sim/MOAICoroutine.cpp
+++ b/src/moai-sim/MOAICoroutine.cpp
@@ -162,13 +162,23 @@ void MOAICoroutine::OnUpdate ( float step ) {
 			if (( result != LUA_YIELD )) {
 			
 				if ( result != 0 ) {
-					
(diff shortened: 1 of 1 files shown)


JDaance:
Impressive!

bullno1:
:beer:
Switch Lua for LuaJIT they said, drop-in replacement they said, ABI level compatibility they said.

seebs:
I'd say that, if it works in testing, this is at least better, and would reduce one of the major pains we have now. That would be nice.

halfnelson:
i guess we could add a ticket to reintroduce custom stack dumping later on (at a lower priority)

PR #956, opened 2014-07-20
Fix MOAITextBox.revealAll not updating spooler

criptych:
Fixed `MOAITextBox.revealAll` not updating the current spooler position, as
`setReveal` does.


criptych:
Updated.

patrickmeehan:
Sounds like I need to think a bit more about expected behavior.

Since @halfnelson has proposed a workaround, I am going to table this one.

@criptych Will you be satisfied using the workaround for now? Happy to revisit this down the road.

criptych:
Sorry for the late reply:  yes, the workaround seems to have the desired effect; thanks @halfnelson.  It's also possible I really just misunderstood the purpose of the function.

PR #958, opened 2014-08-08
Fix `GetMACAddress` for Windows

criptych:
Fixes `ZLAdapterInfo::GetMACAddress()` for Windows systems, used by
`MOAIEnvironment.getMACAddress()` (previously returning zeroes).


halfnelson:
Cheeky use of UUIDs :+1:

PR #959, opened 2014-08-18
Added MOAIBox2DWorld:addRevoluteJointLocal

fivevolthigh:
As discussed on the getmoai.com/forums. Adds a variant function to MOAIBox2DWorld that adds a revolute joint build on local anchors to the relevant A & B bodies, instead of a single, world based anchor as per existing addRevoluteJoint.


halfnelson:
Great stuff. Thanks

PR #960, opened 2014-08-18
fix bug in MOAILuaState::GetLuaObject()

Vavius:
This is from my liquid fun branch. I was doing some heavy table parsing there (tons of physical parameters). 
This fix is for retrieving MOAILuaObject from table by string key.


halfnelson:
Certainly looks like a cut and paste error.

PR #964, opened 2014-08-22
MOAITransformBase.getWorldRot() and .getWorldDir() are broken

DanielSWolf:
`ZLMetaAffine2D/3D.GetRot()` pass their arguments to `atan2()` in the wrong order. Consequently, their results are always inverted and offset by 90°.

`ZLAffine3D.GetHeading()` should apparently return the vector (1, 0, 0), rotated by the rotation part of the transform. In what appears to be a copy-and-paste error from the same method in `ZLAffine2D`, the z coordinate is given the same value as the y coordinate, yielding strange results.

### DIFF
--- a/src/zl-util/ZLAffine2D.h
+++ b/src/zl-util/ZLAffine2D.h
@@ -78,7 +78,7 @@ public:
 	//----------------------------------------------------------------//
 	TYPE GetRot () const {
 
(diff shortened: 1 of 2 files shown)


halfnelson:
This certainly looks like a typo, unless there is some other reason to have them swapped around. 
With the fix, a simple transform with z rotation of 10deg gives a world rot of 10deg, which is as expected. Good find!

PR #966, opened 2014-09-01
Support for WebP image format

DanielSWolf:
Adding support for Google's WebP image format. See discussion at http://getmoai.com/forums/reducing-app-size-t2782/


halfnelson:
It seems to be failing android build for me, missing cpu-features.h

halfnelson:
looks like i just need to link against cpufeatures on android and add android/cpu-features to includes

DanielSWolf:
Makes sense. I didn't realize that cpu.c contained platform-specific code.
Seems like I'll have to set up an Android build environment on my machine after all. This might take a few days, though -- I don't have much time at the moment.

halfnelson:
I have been having a go at this, and managed to get cpufeatures linked in on android build and ended up having to modify the cmakelists.txt to build the  neon optimised c files with -vfpu=neon which was a little tricky with cmake. I haven't tested on linux or ios or osx yet.
You can follow work in progress here: https://github.com/moaiforge/moai-sdk/tree/pull_webp

halfnelson:
Test on actual android device works fine. Works on linux too.

DanielSWolf:
I've merged @halfnelson 's  Android changes into my pull request. It would be great if someone could try it on iOS, since I don't own an Apple device. Otherwise, I think this should be good to go.

patrickmeehan:
Awesome contribution. Unfortunately, the PR went stale. I will close for now, but re-send against develop when/if you have a moment. If not I will bring it in my hand. :)

Also, @DanielSWolf: I have been blissfully ignoring endianess issues. It's sloppy; I'm lazy. Though I don't really have a little-endian system to test on...

PR #968, opened 2014-09-02
Attempt to fix MOAIImageTexture reallocated on each new glyph rendered

Vavius:
Trying to fix `MOAIImageTexture` trashing it's glTexId every time a `InvalidateAll()` is called. 

This happens because image texture is using `MOAIGfxResource::Load()` to force update - after the call it will be in `STATE_PRECREATE`, so on next render step `OnCreate()` is called. Problem is, `Load()` destroys texId and `OnCreate()` will always follow the `CreateTextureFromImage()` path instead of just `UpdateTextureFromImage()`. 

This fix is an update of `MOAIGfxResource` to add state switching method without calling `OnInvalidate()` (which releases texId). 

In attached text case you can see how log is spammed with TEXTURE +/- events every time a new glyph is added to dynamic cache.


halfnelson:
Thanks for this

PR #975, opened 2014-09-21
curl: Fix test for PPC

zeldin:
The tests for PowerPC in `3rdparty/curl-7.19.7/include/curl/curlbuild.h` use preprocessor symbols (`__ppc__` and `__ppc64__`) which are obsolete.  This results in a FTBFS.

The symbols defined by current gcc are `__powerpc__` and `__powerpc64__`.  Note that both are defined on a 64-bit system.

### DIFF
--- a/3rdparty/curl-7.19.7/include/curl/curlbuild.h
+++ b/3rdparty/curl-7.19.7/include/curl/curlbuild.h
@@ -528,7 +528,7 @@
 /* ===================================== */
 
 #elif defined(__GNUC__)
-#  if defined(__i386__) || defined(__ppc__)
+#  if defined(__i386__) || defined(__ppc__) || (defined(__powerpc__) && !defined(__powerpc64__))
 #    define CURL_SIZEOF_LONG           4
 #    define CURL_TYPEOF_CURL_OFF_T     long long
 #    define CURL_FORMAT_CURL_OFF_T     "lld"
@@ -537,7 +537,7 @@
 #    define CURL_SIZEOF_CURL_OFF_T     8
 #    define CURL_SUFFIX_CURL_OFF_T     LL
 #    define CURL_SUFFIX_CURL_OFF_TU    ULL
-#  elif defined(__x86_64__) || defined(__ppc64__)
+#  elif defined(__x86_64__) || defined(__ppc64__) || defined(__powerpc64__)
 #    define CURL_SIZEOF_LONG           8
 #    define CURL_TYPEOF_CURL_OFF_T     long
 #    define CURL_FORMAT_CURL_OFF_T     "ld"


halfnelson:
Thanks for the patch!

I couldn't find any documentation that `__ppc__` and `__ppc64__` are obsolete, and they are in use throughout the current master branch of gcc's source. Unfortunately this is curl's code and their master branch still has this error.

 The guys on XBMC had a similar problem ( http://trac.xbmc.org/ticket/6030 ) and fixed it by using `__powerpc__` and `__powerpc64__` as you did, but also hit a snag that using clang `__powerpc__` **AND**  `__powerpc64__` were both defined (which kind of makes sense).

 I am reluctant to make too many changes to our third party code as it makes it difficult to merge with updated curl down the track.

 Maybe if this could be submitted to the curl guys, then we could patch it with the current version and when we update to latest curl, we still get the fix (since their code will have it now)

zeldin:
> but also hit a snag that using clang powerpc AND powerpc64 were both defined (which kind of makes sense)

Yes, that is what I said.  (It's the same in gcc.)  That is why the patch tests for `defined(__powerpc__) && !defined(__powerpc64__)`.

As for upstreams curl, doesn't that use the tests in configure.ac to create a host specific `curlbuild.h` from `curlbuild.h.in` rather than relying on hardcoded definitions?

If you like, I could make a patch to run the configure script and use the generated `curlbuild.h` instead.

halfnelson:
Ah didn't see the !defined check on my phone (it had wrapped) good stuff. For the other platforms, the moai devs have created a platform specific include folder (eg https://github.com/moai/moai-dev/tree/master/3rdparty/curl-7.19.7/include-osx/curl ) which is includes before the default curl include folder, if you make one of these for your platform (by running configure) it should work fine)  I agree that ideally the cmake scripts should run configure (or actually use the provided cmakelists.txt), but in the short term, an include-linux folder with the fix applied is probably enough.

zeldin:
Yeah, I just noticed that there is support to create the header file from CMake as well, just that the CMakeLists.txt that makes it happen is missing in your repo... :-9

As you say, racking up more include directories doesn't seem like a long term solution.  Currently the `include` directory is used for the platforms "Windows" and "Linux", with #ifdefs inside to separate the actual cases.  I don't really see a point in making a separate `include-linux` directory at this point (there are 28 different arches supported by Linux anyway, some of them with subarches), better to leave it like that until a real solution is integrated.

halfnelson:
the main reason for using a new include-linux directory is to allow us to
have a hand modified header file that is not overwritten when updating to
later versions of curl. Maybe it is time to update to latest curl anyway
and port over to it's cmake build (like we did with SDL)

zeldin:
Beware that the latest curl version contains a warning that the CMake build system is poorly maintained though...

Anyway, I'm going to take a stab at building 7.19.7 with the upstreams CMake files, and see where that takes me.

halfnelson:
we already have it building via CMake, maybe we can just pinch the
configure script part

zeldin:
Considering the state of disarray the upstreams cmake files seems to be in (it's fixable, but then you have the merge issue again), that may indeed be the better option.
I'll see what I can do.

zeldin:
Superseded by #976

Issue #978, opened 2014-09-30
MOAIFont:preloadGlyphs causes crash when a valid font has not been loaded


sclark39:
I am trying to run flower-samples but Moai is crashing. I tracked it down to this bit of script

```
function Font:init(path, charcodes, points, dpi)
    self:load(path)
    self.path = path
    self.charcodes = charcodes
    self.points = points
    self.dpi = dpi

    if charcodes and points then
        self:preloadGlyphs(charcodes, points, dpi)
    end
end
```

The path given doesn't point to a font that I actually have, and when preloadGlyphs is called, the moai exe crashes. Seems like it should error out in this case instead.

halfnelson:
thanks again!

PR #984, opened 2014-10-11
MOAIFont crash fix, zl-gfx assert fix, and fix for silently failing to load the main script

sclark39:
Fixes  moai.exe crashes when attempting to use an unloaded font ( #978 )
Fixes a debug-only assert in zl-gfx due to improper use of an std::string
Makes the moai.exe host log an error when the main script couldn't be loaded instead of silently failing


halfnelson:
looks good to me +1

PR #985, opened 2014-10-11
Make zl-vfs able to load an archive that has had data prepended to it

sclark39:
This fix makes it so zl-vfs can load fused exes that have been created by running 'copy /b moai.exe+src.zip game.exe', rather than requiring a more complicated script ( http://getmoai.com/forums/demo-package-game-in-one-executable-file-t2614/ )

Comment from physfs, from which this approach was inspired:

>  For self-extracting archives, etc, there's crapola in the file
>  before the zipfile records; we calculate how much data there is
>  prepended by determining how far the central directory offset is
>  from where it is supposed to be (start of end-of-central-dir minus
>  sizeof central dir)...the difference in bytes is how much arbitrary
>  data is at the start of the physical file.


patrickmeehan:
How cool is that? Pretty cool.

sclark39:
:)

PR #991, opened 2014-10-18
Fixes for windows build and updated cmake build

halfnelson:
All the fixes needed to allow create_projects_win.bat to create a buildable and runnable moai


halfnelson:
Builds cleanly on linux now

PR #1007, opened 2014-10-27
Regression in font glyph (imagetexture) invalidation

halfnelson:
fixes regression in image texture routines. 
To reproduce run hello moai on android. lock screen, unlock screen, fonts are trashed.

doing nothing on onInvalidate caused the glyphs to be trashed when android was suspended/unsuspended, 

calling load was causing a texture reload every glyph (leak), 

Instead use the invalidateContents method to put the texture in the correct state to load its image into gpu next call (and reuse the texture handle) and leave onInvalidate handling to manage the android suspend/resume

This was working in 1.5-stable

### DIFF
--- a/src/moai-sim/MOAIImageTexture.cpp
+++ b/src/moai-sim/MOAIImageTexture.cpp
@@ -25,7 +25,7 @@
 */
 int MOAIImageTexture::_updateRegion ( lua_State* L ) {
 	MOAI_LUA_SETUP ( MOAIImageTexture, "U" )
-	
+
 	if ( state.GetTop () > 1 ) {
 		ZLIntRect rect = state.GetRect < int >( 2 );
 		self->UpdateRegion ( rect );
@@ -55,7 +55,7 @@ bool MOAIImageTexture::IsValid () {
 //----------------------------------------------------------------//
 MOAIImageTexture::MOAIImageTexture () :
 	mStatus ( INVALID ) {
-	
+
 	RTTI_BEGIN
 		RTTI_EXTEND ( MOAITextureBase )
 		RTTI_EXTEND ( MOAIImage )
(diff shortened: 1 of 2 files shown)


patrickmeehan:
I know this cause a regressing, but in looking at the code I can't see why. Can someone repro the old way and fill me in?

Invalidate () should just call OnInvalidate () then leave the state in STATE_PRELOAD. Load () should just call OnLoad () then leave the state in STATE_PRECREATE. On\* in image texture do nothing.

InvalidateContents () just sets the state directly to STATE_PRECREATE.

I want to go back to the Invalidate ()/Load () idiom as I feel that (barring some renames that need to happen) is more correct (it calls the virtual methods and is aware of the other states).

In my test, it worked. Not sure why it is crapping out on Android.

Will investigate here, but any info would be appreciated. It _should_ work.

halfnelson:
overriding onInvalidate means that when android suspends and invalidates all textures (AKUReleaseGfxContext),  image textures  (including glyph cache) don't get their mTexID reset ( this->mGLTexID = 0; in moaitexturebase oninvalidate), this means on resume, the texture isn't reloaded and the glTexID handle is invalid leading to white blocks instead of glyphs.

so we can't override onInvalidate.

We removed the override in 1.5 and removed the call to invalidate() before the load() in the update region functions. This sort of worked except now every glyph added to the glyphcache was causing the texture to be reloaded. What we really wanted was a way to get to pre_create state without trashing the existing texture handle and texture memory, there didn't seem to be a clean way to do this so vavius added the invalidateContents which just sets pre_create. Just a way of saying everything is happy just copy the image data next time, without having a bunch of deletes or releases happening.

That was our thought process anyway, there might be a cleaner way to solve the problems

patrickmeehan:
Yeah, I just realized the same. Any reason we can't just set mGLTexID = 0 in the override? I made that changed and seems to work right in both use cases.

halfnelson:
won't that recreate a texture each time a glyph is added (updateregion) and
leak texture ID's, (there were no release texture calls to gl at this point)

patrickmeehan:
Yeah. Now that I think about it, I'm not sure why the image texture invalidates at all... since we are re-using the same texture and not re-creating it. So... color me derp. I will re-fix and push again.

PR #1008, opened 2014-10-28
Refactored stream adapters; fix for MOAIImageTexture invalidation

patrickmeehan:
The stream refactor bears some attention. I wrote a sample to test it, but there are probably more test cases. Will eventually adapt that to run as an automated test.

The image texture invalidation is an alternative to InvalidateContents. The intent is to work through the usual gfx resource lifecycle but just to suppress the memory reporting when the texture gets recreated. (Though maybe it _shouldn't_ get suppressed.)


patrickmeehan:
Brain cramp on the whole MOAIImageTexture thing. That last commit should make everything work correctly. My mistake was not implementing the update in the OnBind () method and trying to hijack OnCreate () to do its work.

Issue #1009, opened 2014-10-29
Reopen #998


makotok:
See #998.

MOAITextLabel:getBounds is returning the boundaries of the text. 
MOAITextLabel:getDims is returning the width and height of the text. 
MOAITextLabel:inside and determines whether it is in the text. 

In either case, you will need to perform a forceUpdate function, returns the old value. 
Also, a big change from V1.6. 
If you want to get the size, it is no longer able to be handled in the same way as a MOAIProp.

Vavius:
I see two ways: 
1) getBounds, getDims, inside uses text designer width or height or both based on rectLimits setting. If no limits specified then fallback to stringBounds
2) Respect boundsOverride from MOAIProp

Using larger bounds from the whole text rect can cause unnecessary draw calls. If textLabel will respect boundsOverride we can keep using stringBounds as default bounds. Old behavior can be achieved in Lua with getRect (replacement for getBounds, getDims). If inside() is needed, then you set bounds override.

makotok:
Thank you. I came up with how to fix. 
Take a look because it later pull request.

halfnelson:
pull request is done, is this issue resolved now @patrickmeehan

PR #998, opened 2014-10-23
Fixed the MOAITextLabel.

makotok:
- Behavior getDims, getBounds, the inside function is Fix in the same
  way as V1.5.
- Fixed a problem when changing the size, layout is not reflected
- To maintain compatibility with V1.5, added getString function.

PR #1012, opened 2014-10-30
Fixed the value is not returned problem after calling MOAITextLabel:getDims.

makotok:
See #1009 #998


patrickmeehan:
Thanks, Makoto. But let's hold off on merging this. I want to think about this a little more and talk with you and the others first.

I don't think using bounds override is the way to go, though. That is meant for the end user. We should decide if we care about how the text gets laid out in model space and fix it there.

makotok:
OK. 
The time being, I keep the compatibility with Lua code. (extended Label)
For now, there is no problem even if they are not merged.

seclorum:
Just wanted to report in that, while testing 1.6 changes on Halfnelsons branch today, getBounds() is no longer returning valid xMax/yMax values.  This breaks Hanappe, in case it is of interest in the discussion. 

EDIT: that is to say, Hanappe depends on the prop:getBounds() call working equivalently whether its a Text prop or otherwise..

patrickmeehan:
OK. I looked over the issues a bit. Here's how I want to handle this:

GetBounds () is poorly named and confusing in that the 'bounds' are used for a two different purposes (culling and hit testing). In addition, they can be overridden.

In the case of text label, 'bounds' _should_ mean the bounds for culling, which is the same meaning as the prop bounds: it's the tightest fitting bounds that contains the text, for the most efficient view culling and hit testing.

The text label _rect_ is a distinct concept from prop _bounds_. The _rect_ is a limit used by the layout engine. But the naming is confusing: you could setBounds () and not affect the rect or setRect () and not affect the bounds.

The 'right' fix is to clarify these concepts through better naming. Not that I know what that naming should be. getCullingBounds (), setCullingBounds ()? setLayoutFrame ()? No clue.

In the mean time, as a nod to back compat, I think the right way to go is to use the layout rect as the bounds if it's been set. I should also overload prop's setBounds () to work as an alias for setRect () so that setBounds () and getBounds () will be symmetrical.

If we're all happy with the way that works, I will leave it alone for the rest of 1.x.

patrickmeehan:
OK, this should be addressed by 0e5a073 in #1038. Please take a look and let me know what you think.

PR #1015, opened 2014-11-04
added sdk packaging util to moaiutil

patrickmeehan:


halfnelson:
Awesome stuff. I have a branch with host templates (hand crafted, not cmake) that compile against binary libs for 
- ios (arm,armv7s,arm64)
- osx app store
- vs2012
- android (gradle, android studio project)
- and linux

It also has scripts to generate/compile/export all the libs (universal for ios/osx, combined Moai.LIB for vs2012) and headers for the host templates to use. All libs and hosts can compile and run the samples and all seem to work ok on latest develop branch.

These might be useful. If you have already done this, let me know and I will stop wasting time. Otherwise I can refactor to use the same moai-sdk folder layout and util scripts as your ios/osx hosts

PR #1021, opened 2014-11-08
a bit of everything

ghost:


3rdparty/curl-7.19.7/lib/ssh.c
```diff
@@ -117,9 +117,9 @@
 
 /* Local functions: */
 static const char *sftp_libssh2_strerror(unsigned long err);
-static LIBSSH2_ALLOC_FUNC(libssh2_malloc);
```
patrickmeehan, line 120: This might be a good candidate to fix with header injection, like we do with zl_replace. Depending on where libssh2_malloc is defined.

I didn't have time to look at my_libssh2_malloc, etc. What's the intent? We already hook malloc and free. Is it something we could address there?

ghost, line 120: http://getmoai.com/forums/1-5-devel-curl-libssh2-conflict-t2884/

halfnelson, line 120: This only shows up in linux because the cmake script was changed to run ./configure, the generated curl_config.h does not match the static ones we have already generated for other platforms. The real fix would be to add `–without-libssh2` to the curl configure command line in the curl/cmakelists.txt file under cmake/thirdparty. This would remove dependency on libssh2 being installed on linux boxes where the game is distributed to and bring it in line with the other platforms libcurl support.

ghost, line 120: It works
THX :)

PR #1023, opened 2014-11-11
Add support SDL joysticks + support mouse wheel event; add LINUX c++11 c...

ghost:
...ompliance


ghost:
Example: mouse wheel:

```
MOAIInputMgr.device.mouseWheel:setCallback(function(value) 
        print("value:", value)
    end
)
```

PR #1029, opened 2014-11-15
Memory leak: malloc never released. Is replaced by: alloca

ghost:
Or add free(..) to the end.


patrickmeehan:
Excellent, thanks!

PR #1036, opened 2014-11-29
Proposal: fix assertion 'check_bin'

ghost:
I'm not quite sure. 
What I have seen is that some numbers like 128 was wrong.
For example: 128 ( 10000000 ) replace it with 00000001
But I do not understand exactly what you are trying here.

### DIFF
--- a/src/zl-util/ZLBitBuffer.cpp
+++ b/src/zl-util/ZLBitBuffer.cpp
@@ -33,7 +33,7 @@ bool check_bin ( const void* buffer, cc8* str ) {
 		}
 		bin.append ( byteStr );
 	}
-	return ( strcmp ( bin.c_str (), str ) == 0 );
+	return ( bin.compare( str ) == 0 );
 }
 
 //----------------------------------------------------------------//
@@ -53,7 +53,7 @@ bool check_hex ( const void* buffer, cc8* str ) {
 		hex.push_back ( encode [ byte & 0x0f ]);
 		hex.push_back ( encode [( byte >> 4 ) & 0x0f ]);
 	}
-	return ( strcmp ( hex.c_str (), str ) == 0 );
+	return ( hex.compare( str ) == 0 );
 }
 
 //----------------------------------------------------------------//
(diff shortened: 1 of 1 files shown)


src/zl-util/ZLBitBuffer.cpp
```diff
@@ -414,11 +414,11 @@ void ZLBitBuffer::Test () {
 	
 	memset ( buffer, 0, BUFF_SIZE );
 	Blit ( buffer, 3, srcBuffer, 8, 23, 1 );
-	assert ( check_bin ( buffer, "00010011001110011000000111000000" ));
+	assert ( check_bin ( buffer, "00010011001110010000000111000000" ));
```
patrickmeehan, line 417: Sorry - commented in your repo by accident.

I brought this over into my codebase and it triggered the assert. Did the original _not_ trigger the assert for you?

patrickmeehan:
I am thinking of bits in a bitmap image laid out left to right like:

left                                     right
0 1 2 3 4 5 6 7    8 9 a b c d e f
x=0                                    x=15

So the leftmost pixel corresponds to the first bit. Doing it that way seemed more intuitive for storing visual information and saved me the trouble of having to reverse the bits in each byte.

The ad hoc encode/decode methods in this file are intended to correspond to a _visual_ interpretation of the bits, not a numerical interpretation. So if I write:

00000111 11100000

Then do a blit two pixels to the left I should see:

00011111 10000000

For testing purposes this is easier to check visually than:

11100000 00000111
11111000 00000001

Since the effect on the bitmap image is to shift two pixels to the left, smoothly crossing byte boundaries.

src/zl-util/ZLBitBuffer.cpp
ghost, line 417: The original triggers an assertion. Only the lines shown in the commit, the others are fine.

ghost:
Thanks for the explanation.
Now I close because it commit is meaningless.
I open an issue #1040 instead.

PR #1047, opened 2014-12-23
Myriad fixes; new MOAITrace class; removing moai-harness

patrickmeehan:


david-pershouse:
does the zl_socket code added work on windows. I see a zl_usocket.cpp but the contents of that file don't seem to line up with zl_wsocket.cpp

halfnelson:
working on zl_wsocket.cpp now using zl_usocket as a guide. Almost done

PR #1049, opened 2014-12-23
Linux build: cmake remove all references to wsocket

ghost:


halfnelson:
probably want to remove it from more than just linux, There was an earlier pull request which solves the same problem. Thanks for the contrib though

PR #1054, opened 2014-12-28
Additional ease types (inspired by those found in many other game/javascript frameworks)

Vavius:


src/moai-sim/MOAIEaseType.h
```diff
@@ -28,6 +28,21 @@
 	@const	SOFT_EASE_OUT					t ^ 2
 	@const	SOFT_SMOOTH						t ^ 2
 	@const	SOFT_SMOOTH_EASE_OUT			t ^ 2
+	@const	SINE_EASE_IN
```
DanielSWolf, line 31: It would be great if you could add descriptions to your ease types, like for the existing ones.

src/zl-util/ZLInterpolate.cpp
```diff
@@ -162,6 +164,154 @@ float ZLInterpolate::Curve ( u32 mode, float t ) {
...
+		//................................................................
+		case kBounceSmooth:
+			
+			if ( t < 0.5f ) {
+				return 0.5f * Curve ( kBounceOut, t * 2.0f );
+			}
+			else {
+				return 0.5f + 0.5f * Curve ( kBounceIn, 2.0f * t - 1.f );
+			}
+
```
DanielSWolf, line 237: `BOUNCE_SMOOTH` stops and re-accelarates in the middle:
bounce_smooth
Is this on purpose? I believe it would look better if the middle part was one smooth motion, as in `ELASTIC_SMOOTH`.
elastic_smooth

DanielSWolf:
Great contribution! Those new ease types are much more organic than the existing ones. I especially like the `BOUNCE_*` and `ELASTIC_*` ease types.
I've extended the `anim-curve-draw` sample to show off your new ease types, too. I'll be sending you a pull request!

src/zl-util/ZLInterpolate.cpp
Vavius, line 237: Valid point. `SMOOTH` stands for in_out and inherits bounceIn and bounceOut general look. Slow start in BounceIn is essential, so in order to fix this issue we should make a new function. 
Don't know if it worth it, since `BOUNCE_SMOOTH` (as well as `ELASTIC_SMOOTH`) looks strange to me and I can't see any practical use for them.

PR #1056, opened 2014-12-28
Remove unnecessary files.

makotok:


halfnelson:
thanks makotok, I think @patrickmeehan  wants to keep these around for now. But I will wait for him to comment

patrickmeehan:
Not sure if anyone is using NaCl any more. It's definitely not something we maintain. I'll clobber these for now. If someone screams we can always bring them back.

PR #1058, opened 2015-01-03
HTML host: Minor fix, updated readme files

DanielSWolf:
`moai_lua.h` was out of sync with `moai.lua`. As a result, the HTML5 host didn't start correctly.


DanielSWolf:
I've performed the changes suggested by @halfnelson.

halfnelson:
Thanks for the work on this

PR #1060, opened 2015-01-05
Keyboard

DanielSWolf:
This pull request brings much more robust keyboard handling.
- It introduces an authoritative list of key codes _for all hosts_. So instead of writing `if key == 338`, you can now write `if key == MOAIKeyCode.UP` and be sure it'll work on all hosts and platforms with keyboard support. Before, that key produced 38 using the HTML host, 338 using SDL, and didn't work at all using GLUT on Windows.
- Wherever possible, the new key codes are backwards compatible: All characters keys are still mapped to their respective ASCII codes.
- The new key codes are based on Microsoft's [virtual key codes](http://msdn.microsoft.com/library/windows/desktop/dd375731), so they should work on any keyboard layout, not just US.
- Text input wasn't really possible until now. Depending on the keyboard layout, the same key can produce dozens of different characters. Existing code hat to hard-code logic for the US keyboard that would fail on any other layout. Now there is an additional keyboard event that gives you actual Unicode characters.
- I've included a sample at `samples/input-text` that shows a fully editable textbox, complete with cursor navigation. It works on any keyboard layout and is fully Unicode-compatible.

editor

This pull request is based on this forum discussion: [[WIP CONTRIB] Proposal for robust keyboard handling](http://getmoai.com/forums/proposal-for-robust-keyboard-handling-t2878).


halfnelson:
Rest looks good. Since this is a change of feature instead of new functionality or bug fixes, I think I will let @patrickmeehan  merge it but LGTM :+1:  Thanks for the contribution!

PR #1070, opened 2015-02-13
Latest from Patrick's fork

patrickmeehan:
Wow, it's been longer than I thought. A few notes:

MOAIVertexBuffer and MOAIIndexBuffer are _gone_. There is now just MOAIGfxBuffer. This has caused some changes to MOAIMesh. See samples/gl-cube, etc. This is paving the way for better representation of OpenGL's flexibility around vertex formats and an easier implementation of GLTF.

I should also mention that VAO support was extended to both iOS and desktop. I've only tested this in OSX, but the intent is that all platforms should use VAOs now.

There was some arcane refactoring done around the way partitions apply masks when doing their spatial queries. They now filter by class type. I wound up liking this better than expecting all MOAIProp subclasses to explicitly set their query masks. I also backed away from the idea of "facets" (at least for now) and instead went with class hierarchy; didn't think we needed dynamic composition of objects at runtime.

Per a discussion with @vavius, I restored the behavior of MOAIAction to update in the same frame when started; I had 'fixed' it before to always defer to the next frame (in keeping with my original design intent and faulty first implementation). The new way runs as soon as it can (as long as it hasn't been manually added to an action that will not be revisited in the current frame), but there is also support for deferral via a new 'defer' flag that may be passed to MOAIAction start () and attach () and MOAICoroutine run ().

We also tweaked MOAIWebView (with help from @quentinfasquel or Plumzi) and made some changes to the iOS sample app. These were tested on 7.1 and 8.1 and seem to handle landscape mode and rotation fine.

Finally, MOAIImage and MOAITexture were updated to use a plug-in style architecture for bitmap loaders. As a bonus, we added libpvr (taken from the pvr SDK; subject to their attribution license) to allow software support for PVR on all platforms.

Stay tuned for more MOAIMesh and GLTF stuff coming soon.


seclorum:
Great stuff  - looking forward to digesting this in the near future.

One question on the PVR integration - is this going to be included in the repo as a submodule, or just a copy?  The reason I ask is that in the past I've tried to build your releases and run into the block that I don't have the PVR SDK available locally.

patrickmeehan:
@seclorum The libpvr source is in there as a copy. We copied it out of the SDK, so you won't need to have your own copy of the SDK to build it. :)

halfnelson:
I left a comment on the zl_usocket commit, it seems to force a variable to NULL just before checking to see if it is null, Was this intended, or were you debugging something that got left in?

patrickmeehan:
@halfnelson The zl_usocket stuff is a mess. It was yanked out of luasocket in haste and is definitely not ready for prime time. So what you're seeing is either a bug or just something unfinished. Will look into it soon.

makotok:
There are many modifications. The merged it okay?

PR #1071, opened 2015-02-17
Binary release scripts/mods for cmake builds 1.6

halfnelson:
replace html file packager python script with a moai util script
added build script to create a binary release with host-templates and cmake compiled binaries
added readme for binary release
fixes to host templates to move schemes to shared, fix to osx app template to remove chipmunk ref


makotok:
Thank you for a great fix.
Can be described how to use the README.md?

PR #1095, opened 2015-03-18
Fix hosts

makotok:
Fixed:
- ios
- osx

TODO:
- android ... EGL is can not be solved.
- windows ... My terminal is not.
- html

@halfnelson
Is it possible fix?


david-pershouse:
yeah, when I finish windows phone 8, I will do a fix for other hosts.

PR #1102, opened 2015-03-31
Build Fixes and One Bug Fis

ldeffenb:
Various changes required to
a) Correct syntax errors in Windows .bat files
b) Eliminate compilation errors with VisualStudio 2013
c) Eliminate compilation warnings with compiling for Android under Windows
d) Change default NDK version from 8 to 9 for EGL support
e) Add missing, but required, headers for Android host files
f) Android seems to have an _assert #define that conflicts with MOAITestManager
g) BugFix MOAILogMgr to not use output string as a format, but specify %s to ZLLog::LogF

### DIFF
--- a/src/moai-android-adcolony/host.cpp
+++ b/src/moai-android-adcolony/host.cpp
@@ -1,6 +1,16 @@
 // Copyright (c) 2010-2011 Zipline Games, Inc. All Rights Reserved.
 // http://getmoai.com
 
+// From moai-box2d's pch.h
+#include <moai_config.h>
+#include <zl-util/headers.h>
+#include <moai-core/headers.h>
+#include <moai-sim/headers.h>
+#include <moai-util/headers.h>
+
+// We need to know what's configured as well?
+#include <host-modules/aku_modules_android_config.h>
+
 #include <moai-android-adcolony/host.h>
 #include <moai-android-adcolony/MOAIAdColonyAndroid.h>
 
--- a/src/moai-core/MOAITestMgr.cpp
+++ b/src/moai-core/MOAITestMgr.cpp
@@ -60,6 +60,7 @@ void MOAITestResult::Push ( lua_State* L ) {
 //================================================================//
 
 //----------------------------------------------------------------//
+#ifndef _assert
 int MOAITestMgr::_assert ( lua_State* L ) {
 	MOAI_LUA_SETUP_SINGLE ( MOAITestMgr, "" )
 	
@@ -69,6 +70,7 @@ int MOAITestMgr::_assert ( lua_State* L ) {
 	}
 	return 0;
 }
+#endif
 
 //----------------------------------------------------------------//
 int MOAITestMgr::_comment ( lua_State* L ) {
@@ -243,7 +245,9 @@ void MOAITestMgr::PushResults ( lua_State* L ) {
 void MOAITestMgr::RegisterLuaClass ( MOAILuaState& state ) {
 	
 	luaL_Reg regTable [] = {
+#ifndef _assert
 		{ "assert",					_assert },
+#endif
 		{ "comment",				_comment },
 		{ "error",					_error },
 		{ "pop_test",				_pop_test },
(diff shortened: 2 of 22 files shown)


makotok:
Thank you Fix!

src/moai-core/MOAITestMgr.cpp
```diff
@@ -69,6 +70,7 @@ int MOAITestMgr::_assert ( lua_State* L ) {
 	}
 	return 0;
 }
+#endif
```
halfnelson, line 73: instead of ifdefing this out, couldn't we just rename _assert to _moaiassert or something. This will mean testmgr will still function on android

src/moai-android-adcolony/host.cpp
```diff
@@ -1,6 +1,16 @@
 // Copyright (c) 2010-2011 Zipline Games, Inc. All Rights Reserved.
 // http://getmoai.com
 
+// From moai-box2d's pch.h
+#include <moai_config.h>
+#include <zl-util/headers.h>
+#include <moai-core/headers.h>
+#include <moai-sim/headers.h>
```
halfnelson, line 8: these were all compiling fine for me on android? why drag in all these headers?

ldeffenb, line 8: Without the changes, build-android gives:

D:\MOAI-Stuff\moai-dev3\src\moai-android-adcolony\host.cpp: In function 'void AKUAndroidAdColonyContextInitialize()':

D:\MOAI-Stuff\moai-dev3\src\moai-android-adcolony\host.cpp:29:23: error: 'MOAIAdColonyAndroid' was not declared in this scope

D:\MOAI-Stuff\moai-dev3\src\moai-android-adcolony\host.cpp:29:43: error: 'REGISTER_LUA_CLASS' was not declared in this scope

make[2]: **\* [libmoai/moai-android/moai-android-adcolony/CMakeFiles/moai-android-adcolony.dir/D_/MOAI-Stuff/moai-dev3/src/moai-android-adcolony/host.cpp.o] Error 1

make[1]: **\* [libmoai/moai-android/moai-android-adcolony/CMakeFiles/moai-android-adcolony.dir/all] Error 2

The issue actually is that nothing has pulled in 
<host-modules/aku_modules_android_config.h> leaving 
AKU_WITH_ANDROID_ADCOLONY undefined.  MOAIAdColonyAndroid.h therefore 
doesn't define the class, but MOAIAdColonyAndroid.cpp attempts to 
implement the undefined class.

It turns out that simply adding 
<host-modules/aku_modules_android_config.h> to MOAIAdColonyAndroid.h to 
get the definition of AKU_WITH_ANDROID_ADCOLONY works.  However, if that 
is defined to 0, then the original issue comes back.

The class implementation in MOAIAdColonyAndroid.cpp should be wrapped in 
#if AKU_WITH_ANDROID_ADCOLONY to finish the configuration of the host 
component.

This needs to be done to each of the android host components.  I'm 
willing to do this and remove the then unnecessary headers from the 
corresponding host.cpp files if this is deemed as the proper solution.

Lynn (D)

ldeffenb, line 8: Replace all MOAIAdColonyAndroid.cpp with host.cpp.  Had the wrong file that was having the issue.  At least, at this point.  It may be that the gradle build under Android Studio may need similar conditional compilation in MOAIAdColonyAndroid.cpp.

ldeffenb, line 8: Well, just checked MOAIAdColonyAndroid.cpp and it is already completely wrappered by #if AKU_WITH_ANDROID_ADCOLONY, but it also has not pulled in <host-modules/aku_modules_android_config.h> prior to referencing AKU_WITH_ANDROID_ADCOLONY.  The configurability of the Android host plug-ins isn't quite up to snuff, IMHO.

halfnelson, line 8: The way the android build works it to always compile all host modules to their static libs, then in the host gradle file you pick which android plugins you want and it will link them in to the binary. There should be no massive IFDEF around the code, that is a legacy thing/regression. The new android host module system is as easy as adding "moai-android-adcolony" to a config line and pressing rebuild in android studio to add moai-android-adcolony support

PR #1104, opened 2015-04-01
Standardize and correct AKU_WITH_ANDROID_xxxxx conditionals

ldeffenb:
I've removed the extra headers I previously added to host.cpp now that I understand a bit more about what comes from where.  Now the class definition header file (MOAIxxxxAndroid.h) #includes aku_modules_android_config.h so that the conditionals are defined.  The class implementation file is now conditional after the #includes so that the conditional is defined before being referenced.  And host.cpp is also fully conditional on the same AKU_WITH_ANDROID_xxxxx condition.

I also fixed one method invocation in MOAIFacebookAndroid.cpp (from SetStrongRef to SetRef) now that it is actually compiling given the corrected conditionals.


halfnelson:
the source files should always be built (to provide a binary release) and not have their contents determined by a conditional. Each module "moai-android-twitter.a" will be linked and the header included (or not) via the conditional in the aku_modules_android which is compiled as part of the host (not libmoai).

So to disable adcolony you would still build the full libmoai including moai-android-adcolony, but your aku_modules_config_android.h would set AKU_WITH_ANDROID_ADCOLONY to 0 and the aku_modules_android would then not include the header, and in your gradle config you would not add "moai-android-adcolony" to the list of modules, this means the build task for the host would not link the module and include any of the jars.

ldeffenb:
Got it!  I'll get the conditionals removed from the files to which I 
added them and re-commit.

patrickmeehan:
@ldeffenb - Sorry you had to spin your wheels! We've been cleaning that part of it up internally. Just about to drop a big pull request with our latest Android build and modules stuff.

PR #1111, opened 2015-04-29
moaiutil: Handle non-absolute path

seebs:
If moaiutil is invoked with a partial path (like "../util/moaiutil"),
it gets confused trying to find MOAI_SDK_HOME. Only try to set
MOAI_SDK_HOME if it's not set, and set it a little more cleverly.

### DIFF
--- a/util/moaiutil
+++ b/util/moaiutil
@@ -1,6 +1,28 @@
 #!/usr/bin/env bash
 INVOKE_DIR=$(pwd)
-MOAI_SDK_HOME=$(dirname "${0}")/../
+find_in_path() {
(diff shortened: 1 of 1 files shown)


halfnelson:
What are the advantages of this approach and not simply this after the assignment

``` bash
MOAI_SDK_HOME=$(cd $MOAI_SDK_HOME; pwd)
```

this would force an absolute path. It is also the method used on other scripts in the bin directory.

seebs:
The main advantage is that if someone has put the path to moaiutil in $PATH and just invoked it as "moaiutil", it still works.

PR #1125, opened 2015-10-09
1.7 develop

halfnelson:
Mega merge from fallguy/develop
working cmake builds for windows/linux
osx and ios working from xcode


halfnelson:
merging since at least this will build whereas develop currently doesn't

halfnelson:
@patrickmeehan  could you please look at the latest commits there were some questions there that might be easy for you to answer.

PR #1136, opened 2015-10-14
Fix copy of files on OSX, add usage() docs to moaiutil, include env.sh toolbelt script

seclorum:
Fixed: Copying of files on OSX is best done with ditto to preserve attribs, safe over-writes, etc.
Fixed: moaiutil blows up on poke (i.e. moaiutil #no args != explode)
Added: Basic usage() for moaiutil tool belt instructions
Added: env.sh for setting up MOAI_SDK_HOME and donning the toolbelt.


util/host/main.lua
```diff
@@ -1,22 +1,3 @@
...
-    
-    Usage: moaiutil host <subcommand> <args>
-    
-    Subcommands:
-      host list - Lists available hosts
-      host init - Creates a template host config file used by subsequent commands
-      host build <hostname> - Creates (if it doesn't exist) in hosts folder, and (re)builds the host named <hostname>. 
-      host run <hostname> - Creates (if it doesn't exists) in hosts folder, (re)builds and runs the host named <hostname>. 
-      host create <hostname> - Creates the host in the hosts folder (removing old host) based on latest config settings.
-    ]])
```
halfnelson, line 16: I think that it is still good to have command usage info, so I don't have to wade through tonnes of irrelevant usage info to find out the command to build a host

util/moaiutil.lua
```diff
@@ -5,11 +5,73 @@ require ( 'http' )
...
+MOAI_CMD        = arg [ 3 ]
+SCRIPT_DIR      = string.format ( '%sutil/%s/', MOAI_SDK_HOME, MOAI_CMD )
+
+function usage()
+  print([[
+    MOAI Utility Toolbelt 
+
+    PREPARE to use the toolbelt:
+        MOAI SDK has a toolbox environment script (env.sh) for use prior to loading the toolbelt:
+            source $MOAI_SDK_HOME/env.sh
```
halfnelson, line 17: env.sh is platform specific and is not really needed (adding sdk/util to the system path is enough) Moaiutil startup scripts are designed to locate the moaisdk dir based on the moaiutil.sh/bat location

util/moaiutil.lua
```diff
@@ -5,11 +5,73 @@ require ( 'http' )
...
+        (Presumably you found it already.)
+    
+    INITIALIZE Project folder:              moaiutil init
+        Run from a clean project folder to initialize a moaiutil-managed set of host projects.
```
halfnelson: doesn't need to be a clean project folder, can be run in existing folder too, command should be `moaiutil host init`

util/moaiutil.lua
seclorum, line 17: env.sh works on Linux and OSX of course, and the purpose to have it set up is so that new users have a single script they can run to just up and running.  MOAI_SDK_HOME is used elsewhere in the system too, not just moaiutil ..

util/host/main.lua
seclorum, line 16: Okay, I just wanted to clean it up and not have copies of the text all over the place .. I could add an arg to usage() so if its nil, it spits out everything, if its "host", it spits out the host usage info, etc.

util/moaiutil.lua
```diff
@@ -5,11 +5,73 @@ require ( 'http' )
...
+            host build <hostname> - Creates (if it doesn't exist) in hosts folder, and (re)builds the host named <hostname>. 
+            host run <hostname> - Creates (if it doesn't exists) in hosts folder, (re)builds and runs the host named <hostname>. 
+            host create <hostname> - Creates the host in the hosts folder (removing old host) based on latest config settings.
+        Example:
+            cd newMOAIProject && moaiutil host create ios && moaiutil host create android-studio && moaiutil host create host osx-app && #etc.
+
+    MAKE Developer Documentation:           moaiutil make-cpp-docs
+        Creates compact documentation from the MOAI C/C++ codebase.
+        Example:    
+            cd newMOAIProject && moaiutil make-cpp-docs
```
halfnelson, line 52: this guy is just a stub atm, would be a good idea to mention moaiutil make-lua-docs though

util/util.lua
```diff
@@ -102,14 +102,20 @@ end
...
 
 	if osx then
-		
+		print("COPY: \"" .. srcpath .. "\" -> \"" .. dstpath .. "\"")
 		-- awful, but until there's a better way...
 		moaiCopy( srcpath, dstpath )
 
 		-- so awful
-		os.execute ( string.format ( 'rm -fr %s', dstpath ))
+		if (false) then
```
halfnelson, line 110: if we skip the rm -rf, can we also skip the moaiCopy above?

util/util.lua
```diff
@@ -102,14 +102,20 @@ end
...
+		print( "Warning! removal operation: " .. string.format ( 'rm -fr \"%s\"', dstpath ))
+		os.execute ( string.format ( 'rm -fr \"%s\"', dstpath ))
+		end
 
-		local cmd = string.format ( 'cp -a %s %s', srcpath, dstpath )
+		-- less awful: ditto
+		--local cmd = string.format ( 'cp -rp \"%s\" \"%s\"', srcpath, dstpath )
+		local cmd = string.format ( 'ditto \"%s\" \"%s\"', srcpath, dstpath )
+		print("COPY COMMAND: " .. cmd)
 		print ( cmd )
```
halfnelson, line 119: can probably get rid of the extra print(cmd)

PR #1140, opened 2015-10-17
MOAIGfxDeviceVertexWriter support for writing indexed primitives and properly batch them

Vavius:
- Fix spine batching when using ffd
- Separate indexed and non-indexed paths in GfxDevice. Non-indexed should be usable as before
- Using u16 for indices in GfxDevice to improve portability
- Optimized buffer writes, now avoids some checks that are already performed by GfxDeviceVertexWriter and virtual call. Overall speed-up is about 15% in synthetic sample.


src/moai-spine/MOAISpineSkeleton.cpp
```diff
@@ -724,20 +724,22 @@ void MOAISpineSkeleton::Draw ( int subPrimID, float lod ) {
...
-					gfxDevice.WriteFinalColor4b ();
-				}
-				
-				gfxDevice.EndPrim ();
+			
+			u32 vtxCount = ( u32 )( mVertices.GetTop () / 2 );
+			
+			gfxDevice.BeginPrimIndexed ( ZGL_PRIM_TRIANGLES, vtxCount, trianglesCount );
+
+			for ( u32 i = 0; i < mVertices.GetTop (); i += 2 ) {
```
halfnelson, line 732: does GetTop() need to be called every loop?

patrickmeehan:
Fixed the conflicts and brought it into the fallguy repo (feature/spine) before realizing I need to remove Spine from the community repo. See my comments on Slack.

PR #1144, opened 2015-10-19
moaiutil cleanup, addition of environment command

seclorum:
General cleanup of the moaiutil tool was done with the goal being complete coverage of host and build targets.

On OSX and Linux it should now be possible to perform the following out-of-the-box commands to get a working MOAI development environment, to be subsequently maintained by the moaiutil tool, for all platforms(*):

$ git clone https://github.com/moai/moai-dev/
$ cd moai-dev
$ util/moaiutil environment && source env.sh
$ mkdir ~/SomeProject && cd ~/SomeProject && moaiutil host init && vi hostconfig.lua
$ moaiutil host create ios && moaiutil host create android-gradle && moaiutil host create html5 #&etc.
$ moaiutil host run ios # Get the app running in simulator, &etc.

Note for both Linux and OSX, the presence of a working binary at $MOAI_SDK_HOME/util/moai is now detected, and if it does not exist, moaiutil will build it and put it in place for subsequent use.  This means that $MOAI_SDK_HOME/util/moai can/should be excluded from the repo.

(*-  untested on Windows)

env.sh is produced by running "moaiutil environment".  env.sh is placed in the root of the moai-dev directory for immediate sourcing by developers who have opened a new shell to work on a MOAI project build/test rig.  Once 'source env.sh' is done, the moaiutil toolset can be used in any path location, and the wider use of env.sh can also be to provide other appropriate ENV vars for use in build-server/CI- integration.

Please Note: I also fixed the use of SPINE guard bool in the CMake subsystem, as this twas borked elsewhere, but nevertheless needed to be fixed for my testing of newly created hosts.  Along those lines, I also removed FacebookSDK.framework from the iOS host template.


.gitignore
```diff
@@ -29,11 +29,15 @@ test/xml-results
...
 cmake/projects/
 
+# Build and release products
+cmake/build/
+release/
+
 bin/api-reference-parser/vendor
 docs/api-reference
 /moai-sdk/
-/lib/
```
seclorum, line 35: I don't remember why I did this.  Can someone confirm if this should be gitignored?

halfnelson, line 35: Lib should be ignored. They are the output of the bin/build scripts

util/util.lua
```diff
@@ -102,7 +102,7 @@ end
 copy = function ( dstpath, srcpath )
 
 	if osx then
-		print("COPY: \"" .. srcpath .. "\" -> \"" .. dstpath .. "\"")
+		--print("COPY: \"" .. srcpath .. "\" -> \"" .. dstpath .. "\"")
 		-- awful, but until there's a better way...
 		moaiCopy( srcpath, dstpath )
```
seclorum: I don't know why we need this moaiCopy - its probably redundant because ditto copies files with full attributes on osx.

patrickmeehan:
@seclorum - you've been a busy dude. this all looks great. I will wait for you or @halfnelson pull it in.

Also, I realized increment-sdk-version is legacy and broken. sdk-verion is the util to use. That is the one the build server uses now. Will send the update to 1.7.0 (and some tweaks to sdk-version) in a moment.

PR #1150, opened 2015-10-22
minor tweaks to osx pito

patrickmeehan:
Some changes to pito. Open to discuss these. My thoughts:

1 - Use the xcode project to build. I maintain this project and work with it every day, so for me it is "the" moai.
2 - Think of the bin/folder as a fallback. So if pito/moai exists, always prefer that. Otherwise use bin.
3 - Think of the bin folder as representing "the" moai. People building custom moai's for their own projects should not put anything under the moai project tree. They should instead submodule moai in their repo and have their own build folder.
4 - Not sure if there is ever a reason to have a custom moai in pito. Since pito offers a fixed, known set of functionality it should be fine for it to use "the" moai for any given release.
5 - If there ever is a reason, users can copy moai into pito. Or maybe use a pito command to do that. Then pito will use their special moai.


ant/libmoai-clean.sh
```diff
@@ -3,4 +3,4 @@ set -e # exit on error
 rm -rf libmoai/obj
 rm -rf libmoai/libs
 
-../util/pito ant-libmoai
```
seclorum, line 6: Why this patrick?

seclorum:
Putting a moai bin in the pito folder (maybe we should rename util/ to pito/?) means that no matter what the user does, pito will still continue to work - i.e. if the user blows away their build, there's still a way to run pito independently from what you might do with MOAI builds ..

ant/libmoai-clean.sh
patrickmeehan, line 6: must have been a bad merge. ch had temporarily set some things back to moaiutil so the build server could continue working. must have somehow landed on top of my change back. or I missed it...

seclorum, line 6: That's what I thought - anyway I fixed it and merged yesterday, as you no doubt have noticed.  Moving on ..

PR #1162, opened 2015-10-28
api changes for propForPoint; fixed texture/shader getters; split up iOS build scripts

patrickmeehan:


moai-sdk:
Building IOS simulator Success

PR #1167, opened 2015-11-28
Moai sample browser

halfnelson:
remove some spine references,
restore expat support for html host
fix zip mounting on html host
Sample browser implementation and pito script


seclorum:
Merged, coz I'm on a pito roll ..

PR #1171, opened 2016-04-29
moving alternate build systems and utils to moai-community

halfnelson:
https://github.com/moai/moai-community has the old stuff, and it has been tested with a community release of the 1.7 sdk for all platforms, so this split seems to work well.


patrickmeehan:
Step one in the exciting and fun-filled journey to community 2.0.

PR #1175, opened 2016-04-29
Android binary builds

halfnelson:
The goal of this was to allow the main mk file for android builds able to handle prebuilt static libs (for quick custom host builds), this is done by having an optional "USE_PREBUILT" define that will include a mk file containing the prebuilt_static_library definitions, this file is generated by the ant-libmoai util script.

Also removes hardcoded path for $(MOAI_SDK_HOME)/util/ant-libmoai/ with a global defined by the script for $(MOAI_MODULES)

Splits libraries that have a java module apart so that they can be optionally removed easily in the mk file and not linked (they export jni stuff so are never eliminated by the linker )

These changes shouldn't impact default usage but @joshlytle might want to double check before merge.

### DIFF
--- a/ant/libmoai/jni/Android.mk
+++ b/ant/libmoai/jni/Android.mk
@@ -9,6 +9,7 @@
 	include $(CLEAR_VARS)
 	
 	MOAI_SDK_HOME	:= $(abspath ../../../)
+	
 	MY_ARM_MODE		:= arm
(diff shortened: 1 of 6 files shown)


halfnelson:
@patrickmeehan @JonButt @joshlytle, Does this PR sit okay with you?

PR #1179, opened 2016-06-06
Pull fallguy develop may

halfnelson:
Fallguyconsulting changes up to 3rd June


halfnelson:
I am testing this against with all the prior pull requests in an attempt to catch up. I will post if the tests pan out. (there are a lot of changes)

PR #1193, opened 2017-09-13
Fix inconsistencies in Rocket Lobster tutorial

ryanford:
At some point in the tutorial the constants baseX and baseY become BASE_X and BASE_Y, so I just converted all references to them as the caps variant.

Also, the tutorial never initially creates explosion:main() function. So I changed the reference to it, as if it were being created at that point. 

At the end, the Game Over text doesn't print with a console error that font:getScale() is nil. So tried to consult the docs which used it in the same way. font.getScale like a property seems to get the message to print, at least, maybe that's the correct way to call it?


moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

moai-sdk:
Can one of the admins verify this patch? Type 'test' in a comment to start test.

ryanford:
User error. Disregard

PR #1194, opened 2017-09-28
Feature/community 2.0

halfnelson:
WIP Fixes for 2.0 builds on linux/windows/webassembly, opened pull request for discussion.


patrickmeehan:
@halfnelson Looks good. You said "opening for discussion." I'm fine with pulling this in.

src/moai-sim/MOAIDeck.cpp
```diff
@@ -133,15 +133,14 @@ MOAIDeck::~MOAIDeck () {
 
 //----------------------------------------------------------------//
 bool MOAIDeck::Overlap ( u32 idx, const ZLVec2D& vec, u32 granularity, ZLBounds* result ) {
-
-	return this->Overlap ( idx, vec, granularity, result );
```
david-pershouse: @patrickmeehan did this ever work? am I missing some magic overload logic? this looks like it will always be recursive. is it right to just return false instead? or is there another overlap method that this should be delegating to.

src/moai-sim/MOAIDrawShape.cpp
```diff
@@ -170,7 +170,10 @@ void MOAIDrawShape::DrawPoint ( const ZLVec2D& loc ) {
 //----------------------------------------------------------------//
 void MOAIDrawShape::DrawPoint ( float x, float y, float z ) {
 
-	this->DrawPoint ( x, y, z );
```
david-pershouse, line 173: @patrickmeehan this is the same weird logic where it tries to call itself. I just do nothing here to make the compiler happy, but maybe you know what to do to actually get it to "draw a point"

src/zl-util/ZLBox.cpp
```diff
@@ -225,7 +225,7 @@ void ZLBox::Grow ( const ZLRect& rect, bool first, u32 plane ) {
 
 	ZLBox grow;
 	grow.Init ( rect, plane );
-	this->Grow ( rect, first );
+	this->Grow ( grow, first );
```
david-pershouse, line 228: @patrickmeehan  I assume that this was the correct fix here.

david-pershouse:
I have commented on the lines that would be good to double check. Thanks

src/moai-sim/MOAIDeck.cpp
patrickmeehan: @david-pershouse Looks like I didn't update it to MOAIDeck_Overlap (). I'm bringing in a new naming convention for virtual methods as I refactor/rewrite, but the public-facing methods keep their original names. So a few glitches like this that would result in a recursive call.

src/moai-sim/MOAIDrawShape.cpp
patrickmeehan: @david-pershouse My bad, same as the other. Should be MOAIDrawShape_DrawPoint ().

src/zl-util/ZLBox.cpp
patrickmeehan: @david-pershouse Looks right.

PR #1211, opened 2018-11-05
add moai-osx-expat target for linking .. 

seclorum:
moai-osx-3rdparty-expat target seems to be necessary for ios/osx builds, so I (re-) added it as a target in the Xcode/ project rig.


sshukul (approved):
Works fine for me, thanks!

